Issue #22, opened 2023-02-20
Tenant relationships are ill-defined


JAORMX:
**Describe the bug**

When trying to use the proposed schema in authzed's playground, the tool complains about the `tenant_*` relationships not being defined correctly (the `role` reference is not found). This is because it's missing the prefix.

**To Reproduce**
1. Copy the schema from https://github.com/infratographer/permissions-api/blob/main/internal/spicedbx/schema.go
2. Replace `PREFIX/` with a valid prefix (e.g. `foo/`)
3. Copy it into authzed's playground: https://play.authzed.com/schema

**Expected behavior**
The schema should work out of the box

PR #23, opened 2023-02-20
Add missing prefix in schema

JAORMX:
The tenant definition in the authzed schema had issues. Mainly, the
relationships referring to the roles were lacking prefixes. This fixes
that.

Closes: #22

### DIFF
--- a/internal/spicedbx/schema.go
+++ b/internal/spicedbx/schema.go
@@ -45,10 +45,10 @@ definition PREFIX/tenant {
 	permission edit = editor + parent_tenant->edit_child_tenants
 	permission delete = deleter + parent_tenant->delete_child_tenants
 
-	relation tenant_viewer: role#actor
-	relation tenant_editor: role#actor
-	relation tenant_creator: role#actor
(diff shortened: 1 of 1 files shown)


internal/spicedbx/schema.go
```diff
@@ -45,10 +45,10 @@ definition PREFIX/tenant {
...
 	permission delete = deleter + parent_tenant->delete_child_tenants
 
-	relation tenant_viewer: role#actor
-	relation tenant_editor: role#actor
-	relation tenant_creator: role#actor
-	relation tenant_deleter: role#actor
+	relation tenant_viewer: foo/role#actor
+	relation tenant_editor: foo/role#actor
+	relation tenant_creator: foo/role#actor
+	relation tenant_deleter: foo/role#actor
```
jnschaeffer, line 51: Should these be `PREFIX/role` rather than `foo/role`?

JAORMX: lmao, I commited the wrong thing. Fixed.

PR #29, opened 2023-02-24
Fix CI

JAORMX:
This fixes CI by doing the following:

- Upgrade golangci-lint version
- Fix linting issues
- Fix devcontainer setup
- Add Makefile targets to spawn compose environment through the CLI
- compose: Take container conditions into account
- Use tag v22.2.5 for crdb
- Ensure dev container starts where the repo is mounted.
- Run unit tests in devcontainer


fishnix (approved):
lgtm 👍

PR #48, opened 2023-03-24
Remove invalid API calls from client

JAORMX:
A recent PR (https://github.com/infratographer/permissions-api/pull/47)
simplified the API endpoints and schema. This reflects the change in the
client by removing the API calls that are no longer applicable.

This also removes the explicit setting of the token in favor of using a
ready-to-go http.Client instance such as the one that
https://pkg.go.dev/golang.org/x/oauth2#section-readme may provide.


jnschaeffer (approved):
Nice catch! Thanks for this.

PR #53, opened 2023-03-24
Authenticate API requests

jnschaeffer:
The current implementation of authentication in permissions-api expects a URN in the bearer token instead of a JWT. This PR updates permissions-api's API server to use JWT validation from [`ginjwt`][ginjwt] and URN parsing from [`urnx`][urnx] for more robust and secure authentication of clients making API requests.

Additionally, to make local development easier, a [mock-oauth2-server][mock-oauth2-server] service has been added to the Compose file. This service can be used to get JWT access tokens that permissions-api can be configured to accept instead of using an external identity provider or other authorization server.

[ginjwt]: https://pkg.go.dev/go.hollow.sh/toolbox@v0.5.1/ginjwt
[urnx]: https://pkg.go.dev/go.infratographer.com/x@v0.0.3/urnx
[mock-oauth2-server]: https://github.com/navikt/mock-oauth2-server

### DIFF
--- a/pkg/pubsubx/worker.go
+++ b/pkg/pubsubx/worker.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/nats-io/nats.go"
 	"go.infratographer.com/permissions-api/internal/query"
+	"go.infratographer.com/x/urnx"
 	"go.opentelemetry.io/otel"
 	"go.uber.org/zap"
 	"gocloud.dev/pubsub"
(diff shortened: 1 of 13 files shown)


pkg/pubsubx/worker.go
```diff
@@ -200,15 +201,27 @@ func (s *Subscription) Receive(ctx context.Context, st *query.Stores) error {
 func (s *Subscription) ProcessMessage(ctx context.Context, db *query.Stores, msg *Message) error {
 	switch {
 	case strings.HasSuffix(msg.EventType, ".added"):
-		resource, err := query.NewResourceFromURN(msg.SubjectURN)
+		subjectURN, err := urnx.Parse(msg.SubjectURN)
+		if err != nil {
+			fmt.Println("Error getting resource from URN for subject")
```
fishnix, line 206: why are we `fmt`in' when we have a logger?

fishnix: and below

fishnix (approved):
just the one minor thing, otherwise looks good to me

pkg/pubsubx/worker.go
jnschaeffer: Should be fixed now!

fishnix (approved):

PR #55, opened 2023-03-28
Introduce query engine and schema generation

jnschaeffer:
One of the challenges of permissions management in permissions-api is that we will (likely) need to persist data in two places: SpiceDB for permissions management, and CRDB for other things like update times, role names, and so on. To make this easier in the long run, this PR introduces a few things:

* Namespace: A logical grouping of resource types. All resources in a namespace can refer to each other, but resources between namespaces cannot. Mostly intended for easier testing in SpiceDB
* `query.Engine`: An engine for making authorization queries. The package name might change in the future, but since this is where the authorization logic lives today I figured I would keep it
* `spicedbx.GenerateSchema`: A function for generating a SpiceDB schema based on resource type definitions

API handlers have also been updated to explicitly create and assign roles rather than using the `query` package to do so. This logic needs some refining, however, which will happen as this PR matures.

This PR does not currently have the following:

* [x] `query.Engine` tests
* [x] `spicedbx.GenerateSchema` tests


internal/spicedbx/schema.go
```diff
@@ -1,39 +1,86 @@
...
+{{range .TenantActions}}
+    relation {{$typeName}}_{{.}}_rel: {{$prefix}}/role#subject
+{{- end}}
+{{range .TenantActions}}
+    permission {{$typeName}}_{{.}} = {{$typeName}}_{{.}}_rel + tenant->{{$typeName}}_{{.}}
+{{- end}}
 }
+{{end}}
+`))
+)
```
JAORMX, line 45: Should we already refer to the prefix as a namespace? This way it's easier to read and interpret that it's the intention.

internal/query/tenants.go
```diff
@@ -161,126 +152,30 @@ func builtInRoles(res *Resource) []*pb.RelationshipUpdate {
...
+			Relationships: []types.ResourceRelationship{
 				{
-					Name:       "Parent tenant",
-					Field:      "parent_tenant_id",
-					DBTypes:    "tenant",
-					DBRelation: "parent",
-					Optional:   true,
+					Name:     "tenant",
+					Type:     "tenant",
+					Optional: true,
```
JAORMX, line 163: Given that the subject resource is being deleted from here, do you have a notion of how we'll register subjects? Would we auto-create them as role assignments happen?

jnschaeffer: This got me thinking - RESTful management of resources may not necessarily be what we want here. Resources in SpiceDB are defined only in terms of their relationships, so it's not possible to create a resource that doesn't have relationships.

It may be better to rework this so that the resource API doesn't manage resources, but rather manages roles and relationships.

JAORMX (dismissed):

internal/query/tenants.go
```diff
@@ -74,84 +98,51 @@ func checkPermission(ctx context.Context, db *authzed.Client, req *pb.CheckPermi
 	return ErrActionNotAssigned
 }
 
-func dbRoleName(role string, res *Resource) string {
-	return fmt.Sprintf("%s_%s_%s", res.ResourceType.DBType, res.ID, role)
-}
+func (e *Engine) CreateBuiltInRoles(ctx context.Context, context types.Resource) ([]types.Role, string, error) {
```
fishnix, line 101:
```suggestion
// CreateBuiltInRoles generates the builtin roles for a resource
func (e *Engine) CreateBuiltInRoles(ctx context.Context, context types.Resource) ([]types.Role, string, error) {
```

fishnix (approved):
Makes sense so far to me.  I also started suggesting fixes for things for you that the linter will complain about but didn't do all of them.

cmd/schema.go
```diff
@@ -26,19 +27,32 @@ import (
...
-	Short: "write the schema into SpiceDB",
-	Run: func(cmd *cobra.Command, args []string) {
-		writeSchema(cmd.Context(), globalCfg)
-	},
-}
+var (
+	schemaCmd = &cobra.Command{
+		Use:   "schema",
+		Short: "write the schema into SpiceDB",
+		Run: func(cmd *cobra.Command, args []string) {
```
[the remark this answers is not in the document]
JAORMX: I think it's a good suggestion for a subsequent PR. Can you open a GH issue? Would be a nice low hanging fruit.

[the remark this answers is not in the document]
jnschaeffer: Done! See #57.

fishnix (approved):

PR #65, opened 2023-04-11
Initial proposal for `/allow` endpoint

JAORMX:
This creates an `/allow` endpoint which does the actual permission
check.

It takes three mandatory query parameters:

* `tenant`: The tenant URN
* `resource`: The resource URN
* `action`: The action identifier tag

It also adds a simple OpenAPI v3 spec document.

The intent is to have an easy, general and fairly opinionated endpoint
to do permission checks on. This can be taken programmatically without
much logic in... say... an API Gateway, to do the needed checks without
adding much logic other than gathering the mandatory parameters.

### DIFF
--- a/internal/api/permissions.go
+++ b/internal/api/permissions.go
@@ -20,20 +20,16 @@ import (
 // contain the subject of the request in the "sub" claim.
 //
 // The following query parameters are required:
-// - resource: the resource URN to check
 // - tenant: the tenant URN to check
 // - action: the action to check
+//
+// The following query parameters are optional:
+// - resource: the resource URN to check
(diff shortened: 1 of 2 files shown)


fishnix (commented):
This looks good, I don't have a strong opinion about the previous `has/on` endpoint vs query parameters. I'm curious if we've thought about other ways to provide the tenant, for example in the jwt. Is this particularly untenable in the api-gateway? In this model, how do we account for permission on tenants without resources? do we duplicate the tenant in the resource query field? I could imagine pros/cons for that field being optional.

JAORMX:
@fishnix thansk for the feedback, I think we can address that by making the resource ID optional.

PR #68, opened 2023-04-12
Use Trivy more effectively

jnschaeffer:
The current image scanning workflow leads to Trivy scans being run twice because the test action is run on both PRs and pushes to main. To rectify this, a new security action has been added in this PR that only runs on PRs to scan both the permissions-api Git repository and Docker image. Additionally, repo scanning has been added to the image-build action and a typo in the image tag to be scanned has been fixed.


fishnix (approved):
lgtm! :shipit:

jnschaeffer:
Doesn't look good to me - turns out Trivy caught some things it didn't like in the repo code :angel: It just didn't fail because I didn't have the exit code set.

jnschaeffer:
Decided to disable the `config` scanner on @JAORMX's advice and because [DS0017][ds0017] is yielding false positives, which seems to be a known issue: https://github.com/aquasecurity/defsec/issues/1235

[ds0017]: https://avd.aquasec.com/misconfig/dockerfile/general/avd-ds-0017/

PR #75, opened 2023-04-21
Fix client code and use resource-level authorization

jnschaeffer:
The provided client package was not using the correct authorization HTTP resource. This commit fixes that, and also removes tenant as a query param because we no longer need it to make authorization decisions.

### DIFF
--- a/pkg/client/v1/auth.go
+++ b/pkg/client/v1/auth.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"path"
 	"strings"
 
 	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
--- a/pkg/client/v1/auth.go
+++ b/pkg/client/v1/auth.go
@@ -104,7 +104,7 @@ func newGetRequest(ctx context.Context, uri, endpoint string, query url.Values) 
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/client/v1/auth.go
```diff
@@ -85,23 +89,27 @@ type ServerResponse struct {
...
 }
 
-func newGetRequest(ctx context.Context, uri, endpoint string) (*http.Request, error) {
+func newGetRequest(ctx context.Context, uri, endpoint string, query url.Values) (*http.Request, error) {
 	u, err := url.Parse(uri)
 	if err != nil {
 		return nil, err
 	}
 
 	u.Path = path.Join(apiVersion, endpoint)
```
mikemrm, line 107:
```suggestion
	u = u.JoinPath(apiVersion, endpoint)
```

pkg/client/v1/auth.go
```diff
@@ -85,23 +89,27 @@ type ServerResponse struct {
...
 
+	if len(query) > 0 {
+		u.Path = fmt.Sprintf("%s?%s", u.Path, query.Encode())
```
mikemrm, line 110:
```suggestion
		u.RawQuery = query.Encode()
```

Shouldn't be formatting the string manually. Let url deal with that.

fishnix (approved):

PR #81, opened 2023-04-28
Add NATS consumer for resource lifecycle events

jnschaeffer:
All resource lifecycle events are expected to flow to permissions-api through NATS. This PR introduces a NATS JetStream consumer based on a prior implementation of NATS worker logic in permissions-api that listens for events on resource types.

Additionally, this commit adds NATS-specific development tooling and updates the Makefile to perform some of the tool fetching/building logic that was previously in the dev container Dockerfile to reduce build times in CI.

[rm-doc]: https://github.com/infratographer/permissions-api/blob/11259e74d320bf8f8030133244160fe0db21e34b/docs/resource_management.md

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/signal"
 	"strings"
+	"syscall"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -58,9 +59,9 @@ func init() {
 	rootCmd.PersistentFlags().String("spicedb-prefix", "", "spicedb prefix")
 	viperx.MustBindFlag(viper.GetViper(), "spicedb.prefix", rootCmd.PersistentFlags().Lookup("spicedb-prefix"))
 
-	// Set up SIGINT listener
+	// Set up SIGINT/SIGTERM listener
 	sigCh = make(chan os.Signal, 1)
-	signal.Notify(sigCh, os.Interrupt)
+	signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
 }
 
 // initConfig reads in config file and ENV variables if set.
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"log"
 	"os"
+	"os/signal"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -20,6 +21,7 @@ var (
 	cfgFile   string
 	logger    *zap.SugaredLogger
 	globalCfg *config.AppConfig
+	sigCh     chan os.Signal
 )
 
 // rootCmd represents the base command when called without any subcommands
@@ -55,6 +57,10 @@ func init() {
 	viperx.MustBindFlag(viper.GetViper(), "spicedb.verifyca", rootCmd.PersistentFlags().Lookup("spicedb-verifyca"))
 	rootCmd.PersistentFlags().String("spicedb-prefix", "", "spicedb prefix")
 	viperx.MustBindFlag(viper.GetViper(), "spicedb.prefix", rootCmd.PersistentFlags().Lookup("spicedb-prefix"))
+
+	// Set up SIGINT listener
+	sigCh = make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt)
 }
 
 // initConfig reads in config file and ENV variables if set.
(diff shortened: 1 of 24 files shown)


cmd/root.go
```diff
@@ -55,6 +57,10 @@ func init() {
 	viperx.MustBindFlag(viper.GetViper(), "spicedb.verifyca", rootCmd.PersistentFlags().Lookup("spicedb-verifyca"))
 	rootCmd.PersistentFlags().String("spicedb-prefix", "", "spicedb prefix")
 	viperx.MustBindFlag(viper.GetViper(), "spicedb.prefix", rootCmd.PersistentFlags().Lookup("spicedb-prefix"))
+
+	// Set up SIGINT listener
+	sigCh = make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt)
```
adammohammed, line 63: Worth catching `TERM` too? that's what k8s will actually send when rolling a new version out.

jnschaeffer: Definitely - good catch.

internal/pubsub/subscriber.go
```diff
@@ -0,0 +1,372 @@
...
+	}
+
+	for _, name := range c.resourceTypeNames {
+		subject := fmt.Sprintf("%s.%s.>", c.prefix, name)
+		queueName := fmt.Sprintf("%s-%s", c.consumer, name)
+
+		subscription, err := c.js.QueueSubscribe(subject, queueName, c.receiveMsg, subOpts...)
+		if err != nil {
+			return err
+		}
```
fishnix, line 188: This will process each subscription serially btw - which might be what we want, just pointing it out.

internal/pubsub/subscriber.go
```diff
@@ -0,0 +1,372 @@
...
+	for field, value := range fields {
+		relation, found := strings.CutSuffix(field, fieldRelationshipSuffix)
+		if !found {
+			continue
+		}
+
+		subjResource, err := c.newResourceFromString(value)
+		if err != nil {
+			c.logger.Errorw("error parsing field - will not reprocess", "event_type", eventTypeCreate, "field", field, "error", err.Error())
+			return msg.Ack()
```
fishnix, line 239: all of these buried `Acks` and `Nacks` make me a little uncomfortable - I wonder if there's a better way to handle that?  I realize *why* its this way - ie. how to handle a bad message is different in each case, just a thought.

internal/pubsub/subscriber.go
```diff
@@ -0,0 +1,372 @@
...
+	default:
+		err = c.handleUnknownEvent(ctx, msg, payload)
+	}
+
+	if err != nil {
+		c.logger.Errorw("error handling message", "error", err.Error())
+		return
+	}
+
+	err = msg.Ack()
```
fishnix, line 365: another style comment - i prefer to see shadowed `err` with inline case statements over `err` reassignment ie.

```go
if err := msg.Ack(); err != nil {
...
}
```

☝️ imo its easier to forget a return or other thing on a reassigned error and have bugs that are hard to find bc the var is defined

although I _will_ reassign error if im initializing another var. ie

```golang

foo, err := doThing()
if err != nil {
...
}

bar, err := doOtherThing()
if err != nil {
...
}

**Review by fishnix — DISMISSED** [review]
I just had some small style related comments otherwise this looks good to me! 👍

File: internal/pubsub/subscriber.go [hunk 5]
```diff
@@ -0,0 +1,372 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		c.nc = nc
+	}
+
+	// If we don't have a logger, use a nop logger.
+	if c.logger == nil {
+		c.logger = defaultLogger()
```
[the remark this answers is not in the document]
jnschaeffer: I can do that. I was sort of looking to avoid unnecessary assignments, but it doesn't really matter much either way.

internal/pubsub/subscriber.go
jnschaeffer: Good call-out. I think that's fine for now, since we don't have a good sense for what our resource limits will be or whether certain subjects will require more resources to process than others. Leaving this at a single subscription per resource type is probably better than trying to optimize for things we don't know yet.

internal/pubsub/subscriber.go
jnschaeffer: Addressed this by changing the semantics up a bit - we only `Nak` now if we get an error handling the event and don't consider malformed messages to be errors.

fishnix (approved):

PR #106, opened 2023-06-13
IAPL schema generation

jnschaeffer:
This PR updates permissions-api to use the Infratographer authorization policy language (final name TBD) as defined in this design document: https://gist.github.com/jnschaeffer/29df29ee3fc8aca905cafb45919e9843

The overall goal here is to move towards an external policy being provided to permissions-api, then modified with some resources like roles and subjects that permissions-api needs to do its job. Some work in that realm that is explicitly out of scope for this PR:

* External policy definition and consumption
* Moving from URNs to prefixed IDs for resources
* Moving to new Infratographer message formats

These items will be addressed in subsequent PRs.


internal/iapl/policy.go
```diff
@@ -0,0 +1,300 @@
...
+
+// Condition represents a necessary condition for performing an action.
+type Condition struct {
+	RoleBinding        *ConditionRoleBinding
+	RelationshipAction *ConditionRelationshipAction
+}
+
+// ConditionRoleBinding represents a condition where a role binding is necessary to perform an action.
+type ConditionRoleBinding struct {
+}
```
fishnix, line 56:
```suggestion
type ConditionRoleBinding struct {}
```

internal/iapl/policy.go
```diff
@@ -0,0 +1,300 @@
...
+	ActionName string
+}
+
+// Policy represents an authorization policy as defined by IAPL.
+type Policy interface {
+	Validate() error
+	Schema() []types.ResourceType
+}
+
+var _ Policy = &policy{}
```
fishnix, line 71: This doesn't seem like its doing anything?

internal/iapl/policy.go
```diff
@@ -0,0 +1,300 @@
...
+
+	out := policy{
+		rt: rt,
+		ta: ta,
+		ac: ac,
+		rb: rb,
+		p:  p,
+	}
+
+	return &out
```
fishnix, line 117:
```suggestion
	return &policy{
		rt: rt,
		ta: ta,
		ac: ac,
		rb: rb,
		p:  p,
	}
```

internal/iapl/policy.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+func (v *policy) validateTypeAliases() error {
```
fishnix, line 120: check for a `nil` policy? here and below?

internal/spicedbx/schema.go
```diff
@@ -108,24 +54,205 @@ func GenerateSchema(namespace string, resourceTypes []types.ResourceType) (strin
...
+						TargetTypeName: "subject",
+					},
+				},
+			},
+			{
+				Name:     "user",
+				IDPrefix: "idenusr",
+			},
+			{
+				Name:     "client",
```
fishnix, line 76: whats a client?

fishnix (commented):
Looks like a good start to me - just a few comments so far.

internal/iapl/policy.go
jnschaeffer: It's making sure `policy` implements `Policy`, but that can be removed.

internal/iapl/policy.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+// Relationship represents a named relation between two resources.
+type Relationship struct {
+	Relation       string
+	TargetTypeName string
+}
+
+// TypeAlias represents a named alias to multiple concrete resource types.
+type TypeAlias struct {
```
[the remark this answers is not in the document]
jnschaeffer: Resolving this because we switched to union as the term to use, per the design discussion.

internal/spicedbx/schema.go
jnschaeffer: Client here refers to an OAuth 2.0 client. Once this is more mature we'll have a YAML file that defines the policy instead of hardcoding things here, with the exception of roles.

internal/iapl/policy.go
jnschaeffer: In general I don't like to do nil checks for pointer receivers unless there's a legitimate reason they could be nil. In this case there isn't, so a panic would be the expected result for me if someone tried to call `validateTypeAliases()` on a nil policy.

internal/iapl/policy.go
```diff
@@ -0,0 +1,320 @@
...
+		p:  p,
+	}
+
+	return &out
+}
+
+func (v *policy) validateUnions() error {
+	for _, typeAlias := range v.p.Unions {
+		if _, ok := v.rt[typeAlias.Name]; ok {
+			return fmt.Errorf("%s: %w", typeAlias.Name, ErrorInvalidAlias)
```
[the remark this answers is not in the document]
mikemrm: `%s: invalid union, resource type already defined` may be better

mikemrm (approved):

PR #108, opened 2023-06-16
Update README to refer to correct permissions check route

jnschaeffer:
The README previously referred to the old API route, /has, for permissions checks. This commit updates it to refer to the current route, /allow.


fishnix (approved):

jnschaeffer:
Merging this as-is because the linting errors are about deprecation notices we're actively fixing.

PR #109, opened 2023-06-16
update to use new event handling

mikemrm:
This adds the new x/events handling.

Note:

There are some issues with detecting Ack/Nack's on messages. Which has resulted in some tests not being able to be done that check for Nacked messages.


internal/pubsub/subscriber.go
```diff
@@ -2,367 +2,232 @@ package pubsub
...
+		msg.Ack()
 
-// Stop drains all subscriptions for the client.
-func (c *Client) Stop() error {
-	for _, sub := range c.subscriptions {
-		err := sub.Drain()
-		if err != nil {
-			return err
-		}
+		return err
```
jnschaeffer, line 132: If this branch gets hit then we'll call `msg.Ack()` and return a non-nil error, which will result in `msg.Nack()` being called above. This should probably just log and return nil.

internal/pubsub/subscriber.go
```diff
@@ -2,367 +2,232 @@ package pubsub
...
-		c.logger.Errorw("error acking message", "error", err.Error())
+	switch events.ChangeType(changeMsg.EventType) {
+	case events.CreateChangeType:
+		err = s.handleCreateEvent(ctx, msg, changeMsg)
+	case events.UpdateChangeType:
+		err = s.handleUpdateEvent(ctx, msg, changeMsg)
+	case events.DeleteChangeType:
+		err = s.handleDeleteEvent(ctx, msg, changeMsg)
+	default:
+		s.logger.Debugw("ignoring msg, not a create, update or delete event", zap.String("event-type", changeMsg.EventType))
```
jnschaeffer, line 145: Logging this as a warning seems more appropriate.

internal/pubsub/subscriber.go
```diff
@@ -2,367 +2,232 @@ package pubsub
...
-		return types.Resource{}, err
+		return err
 	}
 
-	return c.qe.NewResourceFromID(id)
+	return nil
 }
 
-func (c *Client) createRelationships(ctx context.Context, msg *nats.Msg, resource types.Resource, fields map[string]string) error {
+func (s *Subscriber) createRelationships(ctx context.Context, msg *message.Message, resource types.Resource, additionalSubjectIDs []gidx.PrefixedID) error {
```
jnschaeffer, line 155: So `additionalSbjectIDs` isn't the abstraction we want here - that's more of a generic "things that are related to this thing" list. I'm pretty sure what we actually want is a `map[string]gidx.PrefixedID` or something similar, where that map is built from `events.ChangeMessage.SubjectFields` or `events.ChangeMessage.Changes` as follows:

* For the given resource, look up its type in the engine schema
* For each field in the map, check to see if a relation exists with that name
* If it does, look up the target types for the relation
* If it does, ensure the prefix in the ID matches at least one prefix in the target types

As a concrete example, say we have a payload that look like so:
```json
{
    "subjectID": "loadbal-abc123",
    "subjectFields": {
        "name": "my-cool-loadbalancer",
        "owner": "tnntten-abc123"
    }
}
```

Then, assume a policy exists (in YAML) like so:
```yaml
resourceTypes:
  - name: tenant
    idPrefix: tnntten
  - name: loadbalancer
    idPrefix: loadbal
    relationships:
      - relation: owner
        targetTypeNames:
          - tenant
```

When we process the event, we should load the resource type definition from the schema. From there, we can see `owner` is a relation on the authorization policy for a `loadbalancer` resource, but `name` is not. Thus, we should persist anything called `owner` as a relationship in SpiceDB, and check that its ID prefix is `tnntten` before we do.

jnschaeffer (changes requested):
Looks decent so far - from what I understand we can't reintroduce the testing logic to check that a message was acked because the NATS server isn't directly exposed in `x/testing/eventtools`. Thus, most of my input is around the code itself.

internal/pubsub/subscriber.go
mikemrm: This makes sense. Right now the services aren't using the `SubjectFields` attribute and is something I think we need to discuss further. I believe once we've figured that out, this will be an easy transition. For now I've published a few changes to handle the `AdditionalSubjectIDs` better.

internal/pubsub/subscriber.go
```diff
@@ -2,367 +2,247 @@ package pubsub
...
+
+	switch events.ChangeType(changeMsg.EventType) {
+	case events.CreateChangeType:
+		err = s.handleCreateEvent(ctx, msg, changeMsg)
+	case events.UpdateChangeType:
+		err = s.handleUpdateEvent(ctx, msg, changeMsg)
+	case events.DeleteChangeType:
+		err = s.handleDeleteEvent(ctx, msg, changeMsg)
+	default:
+		s.logger.Warnw("ignoring msg, not a create, update or delete event", zap.String("event-type", changeMsg.EventType))
```
fishnix, line 153: IMO we should be structuring these with `.` notation instead of `-`, although I haven't looked around to see what's being done elsewhere so feel free to ignore if there's an established pattern.

internal/query/mock/mock.go
```diff
@@ -71,8 +76,28 @@ func (e *Engine) DeleteRelationships(ctx context.Context, resource types.Resourc
 
 // NewResourceFromID creates a new resource object based on the given ID.
 func (e *Engine) NewResourceFromID(id gidx.PrefixedID) (types.Resource, error) {
+	prefix := id.Prefix()
+
+	var rType *types.ResourceType
+
+	if e.schema == nil {
+		e.schema = iapl.DefaultPolicy().Schema()
```
fishnix, line 84: Feels like the schema should be preloaded and this should be an error?

fishnix (dismissed):
couple of minor comments, otherwise lgtm 👍

internal/query/mock/mock.go
mikemrm: This is just for the mock. Most tests just define an empty mock Engine. So this simplifies to use the default schema if we aren't testing something specific.

internal/pubsub/subscriber.go
mikemrm: I believe in most places we actually use `_` which I can update this to use.

fishnix (approved):

fishnix (approved):

PR #123, opened 2023-06-21
add worker http server for health checks

mikemrm:
Adds http server for worker so the chart can properly run health checks.

Additionally updates the health check for spicedb to read the schema for it's health check.


fishnix (approved):
:shipit:

PR #125, opened 2023-06-22
Add permissions middleware

mikemrm:
Produces an echo middleware which adds a checker function into the
request context that is later used to check if the current actor has
access to the requesting resource and action.


pkg/permissions/permissions.go
```diff
@@ -0,0 +1,203 @@
...
+
+		return nil
+	}
+}
+
+// New creates a new Permissions instance
+func New(config Config, options ...Option) (*Permissions, error) {
+	p := &Permissions{
+		enabled:        config.URL != "",
+		client:         http.DefaultClient,
```
fishnix, line 140: let's use a sane client with a really short timeout instead of the default client

fishnix (changes requested):
just the one change - otherwise looks good!

pkg/permissions/permissions.go
fishnix: (unless you think its better for implementers to pass a client - but that feels like a 🦶 🔫

mikemrm: yeah fair point

fishnix (approved):

PR #138, opened 2023-07-12
Add removal endpoints

mikemrm:
This adds removal endpoints for Relationships, Roles and Assignments

### DIFF
--- a/internal/api/assignments.go
+++ b/internal/api/assignments.go
@@ -117,7 +117,7 @@ func (r *Router) assignmentDelete(c echo.Context) error {
 
 	subjResource, err := r.engine.NewResourceFromID(subjID)
 	if err != nil {
-		return echo.NewHTTPError(http.StatusBadRequest, "error deleting assignment").SetInternal(err)
+		return echo.NewHTTPError(http.StatusBadRequest, "error parsing resource type from subject").SetInternal(err)
 	}
 
 	role := types.Role{
--- a/internal/api/assignments.go
+++ b/internal/api/assignments.go
@@ -91,3 +91,47 @@ func (r *Router) assignmentsList(c echo.Context) error {
 
 	return c.JSON(http.StatusOK, out)
 }
+
+func (r *Router) assignmentDelete(c echo.Context) error {
+	roleIDStr := c.Param("role_id")
+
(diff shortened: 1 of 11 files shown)


internal/api/assignments.go
```diff
@@ -91,3 +91,47 @@ func (r *Router) assignmentsList(c echo.Context) error {
...
+	}
+
+	subjID, err := gidx.Parse(reqBody.SubjectID)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "error parsing subject ID").SetInternal(err)
+	}
+
+	subjResource, err := r.engine.NewResourceFromID(subjID)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "error deleting assignment").SetInternal(err)
```
fishnix, line 120:
```suggestion
		return echo.NewHTTPError(http.StatusBadRequest, "error parsing resource type from subject").SetInternal(err)
```

internal/api/relationships.go
```diff
@@ -124,3 +124,54 @@ func (r *Router) relationshipsList(c echo.Context) error {
...
+
+	ctx, span := tracer.Start(c.Request().Context(), "api.relationshipsDelete", trace.WithAttributes(attribute.String("id", resourceIDStr)))
+	defer span.End()
+
+	resourceID, err := gidx.Parse(resourceIDStr)
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "error parsing resource ID").SetInternal(err)
+	}
+
+	var reqBody deleteRelationshipRequest
```
fishnix, line 139: Why not use query params for this instead of a DELETE body?

fishnix (dismissed):
just some minor stuff, otherwise looks good 👍

internal/api/relationships.go
mikemrm: using a query parameter seems odd for a delete request. however having everything in the url path is pretty standard. I could change it to be `DELETE /resources/:id/relationships/:relation/:related_id`. Thoughts about that?

fishnix: i think this is fine for now 👍

fishnix (approved):

PR #146, opened 2023-07-24
Support bulk checks

mikemrm:
This adds support for requesting access to multiple resources and actions.


internal/api/permissions.go
```diff
@@ -72,6 +90,169 @@ func (r *Router) checkAction(c echo.Context) error {
...
+type checkAction struct {
+	ResourceID string `json:"resource_id"`
+	Action     string `json:"action"`
+}
+
+type checkStatus struct {
+	Resource types.Resource
+	Action   string
+	Error    error
+}
```
jnschaeffer, line 106: I think I'd break this up into two structs: A `checkRequest` containing the resource/action pair and a `checkResult` containing the `checkRequest` object plus an error.

internal/api/permissions.go
```diff
@@ -72,6 +90,169 @@ func (r *Router) checkAction(c echo.Context) error {
...
+			Resource: resource,
+			Action:   check.Action,
+		}
+	}
+
+	if len(errs) != 0 {
+		return echo.NewHTTPError(http.StatusBadRequest, "invalid check request").SetInternal(multierr.Combine(errs...))
+	}
+
+	checkCh := make(chan *checkStatus)
```
jnschaeffer, line 172: So looking at this code, a channel of pointers seems risky. IMO we have two options here:
* Spin up goroutines that receive off a `<-chan checkRequest` (see above) and send to a `chan<- checkResponse`
* Use `sync.WaitGroup` and coordinate writes to a slice

The first option is more idiomatic Go, and I think is generally more robust against introducing data races. I'd recommend we move towards that approach. With that in mind, I'd recommend we make this a buffered channel, like `make(chan checkResponse, len(reqBody.Actions))` and pre-populate the channel and close it before doing any work. That way we have a queue already primed for workers to consume from and don't have to do further coordination of work.

internal/api/permissions.go
```diff
@@ -72,6 +90,169 @@ func (r *Router) checkAction(c echo.Context) error {
...
+
+	wg := new(sync.WaitGroup)
```
jnschaeffer, line 174: Rather than use a `sync.WaitGroup` here, we can instead provide a channel for workers to send results to and iterate over that until we get back as many results as we expect, then close the channel to free resources.

Additionally, we can set up a new context using [`context.WithCancelCause`](https://pkg.go.dev/context#WithCancelCause), then when receiving results cancel the entire operation if we get any errors. SpiceDB seems pretty good at handling context propagation and terminating operations early so we can have things fail faster that way.

internal/api/permissions.go
```diff
@@ -72,6 +90,169 @@ func (r *Router) checkAction(c echo.Context) error {
...
+
+		wg.Wait()
+	}()
+
+	select {
+	case <-doneCh:
+	case <-ctx.Done():
+		return echo.NewHTTPError(http.StatusInternalServerError, "request cancelled").WithInternal(ctx.Err())
+	case <-time.After(maxCheckDuration):
+		return echo.NewHTTPError(http.StatusInternalServerError, "checks didn't complete in time")
```
jnschaeffer, line 217: Setting up a context using [`context.WithTimeout`](https://pkg.go.dev/context#WithTimeout) is more idiomatic here and lets us handle all errors in one place.

jnschaeffer (changes requested):

internal/api/permissions.go
mikemrm: makes sense, updated.

internal/api/permissions.go
mikemrm: Removed WaitGroup and handle waiting with channels.

internal/api/permissions.go
mikemrm: Good point, updated.

internal/api/permissions.go
mikemrm: Updated

internal/api/permissions.go
```diff
@@ -72,6 +90,189 @@ func (r *Router) checkAction(c echo.Context) error {
...
+
+				return
+			}
+
+			if count == len(reqBody.Actions) {
+				close(doneCh)
+				close(resultsCh)
+			}
+		}
+	}()
```
jnschaeffer, line 251: Rather than ranging over the channel, if we know how many results we expect back we should be able to just read those. Something like this:

```go
for i := 0; i < numChecks; i++ {
  select {
    case result := <-resultsCh:
      // check for errors
    case <-ctx.Done():
      // context error handling
  }
}
```

Doing this would let us take out `doneCh` completely and remove the need to process results in a separate goroutine from the rest of the handler.

jnschaeffer (commented):
Looking good - more thoughts.

internal/api/permissions.go
```diff
@@ -72,6 +90,189 @@ func (r *Router) checkAction(c echo.Context) error {
...
+
+				// Check if doneCh has been closed, if so, don't write to resultsCh.
+				select {
+				case <-doneCh:
+					return
+				default:
+				}
+
+				resultsCh <- *result
+			}
```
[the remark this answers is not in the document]
mikemrm: ah yeah, took me a minute but I see what you mean. Thanks!

internal/api/permissions.go
```diff
@@ -72,6 +90,189 @@ func (r *Router) checkAction(c echo.Context) error {
...
+	ctx, cancel := context.WithTimeout(ctx, maxCheckDuration)
+
+	defer cancel()
+
+	for i := 0; i < r.concurrentChecks; i++ {
+		go func() {
+			for check := range requestsCh {
+				result := &checkResult{
+					Request: check,
+				}
```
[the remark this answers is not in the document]
mikemrm: ah good call, fixed!

internal/api/permissions.go
mikemrm: ah yes, i see what you mean. it will just populate the errors with the same context error for all the ones which were cancelled. makes sense!

jnschaeffer (approved):
Looks good! Two small things - neither are blocking.

internal/api/permissions.go
```diff
@@ -72,6 +90,174 @@ func (r *Router) checkAction(c echo.Context) error {
...
+
+	if internalErrors != 0 {
+		return echo.NewHTTPError(http.StatusInternalServerError, "an error occurred checking permissions").SetInternal(multierr.Combine(allErrors...))
+	}
+
+	if unauthorizedErrors != 0 {
+		msg := fmt.Sprintf("subject '%s' does not have permission to the requested resource actions", subject)
+
+		return echo.NewHTTPError(http.StatusForbidden, msg).SetInternal(multierr.Combine(allErrors...))
+	}
```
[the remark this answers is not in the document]
mikemrm: will track it and dig into it.

PR #147, opened 2023-07-24
Get role resources

mikemrm:
Adds endpoints to retrieve a role as well as retrieve a roles resource assignment.

### DIFF
--- a/internal/query/relations.go
+++ b/internal/query/relations.go
@@ -572,6 +572,44 @@ func (e *engine) listRoleResourceActions(ctx context.Context, role types.Resourc
 	return resourceActions, nil
 }
 
+// GetRole gets the role with it's actions.
+func (e *engine) GetRole(ctx context.Context, roleResource types.Resource, queryToken string) (types.Role, error) {
+	var (
+		resActions map[types.Resource][]string
+		err        error
+	)
+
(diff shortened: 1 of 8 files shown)


internal/query/relations.go
```diff
@@ -572,6 +572,75 @@ func (e *engine) listRoleResourceActions(ctx context.Context, role types.Resourc
...
+
+		// roles are only ever created for a single resource, so we can break after the first one is found.
+		if len(resActions) != 0 {
+			break
+		}
+	}
+
+	if len(resActions) > 1 {
+		e.logger.Warnw("role is assigned to more than one resource", "role.id", roleResource.ID.String())
+	}
```
jnschaeffer, line 596: IMO it'd probably be better for us to return an error here, since this is something that shouldn't happen.

internal/query/relations.go
```diff
@@ -572,6 +572,75 @@ func (e *engine) listRoleResourceActions(ctx context.Context, role types.Resourc
...
```
jnschaeffer, line 634: Same as above.

jnschaeffer (changes requested):

jnschaeffer (approved):

PR #148, opened 2023-07-27
update to use new events handling

mikemrm:
Updates to use the new [x/events](https://github.com/infratographer/x/pull/130) events package which removes the watermill format as well as implements the new AuthRelationshipRequest and Response events.

### DIFF
--- a/chart/permissions-api/templates/deployment-worker.yaml
+++ b/chart/permissions-api/templates/deployment-worker.yaml
@@ -58,25 +58,46 @@ spec:
             - name: PERMISSIONSAPI_SERVER_TRUSTED_PROXIES
               value: "{{ join " " . }}"
           {{- end }}
-            - name: PERMISSIONSAPI_EVENTS_SUBSCRIBER_URL
-              value: "{{ .Values.config.events.url }}"
-            - name: PERMISSIONSAPI_EVENTS_SUBSCRIBER_TIMEOUT
-              value: "{{ .Values.config.events.timeout }}"
-            - name: PERMISSIONSAPI_EVENTS_SUBSCRIBER_PREFIX
-              value: "{{ .Values.config.events.prefix }}"
-            - name: PERMISSIONSAPI_EVENTS_SUBSCRIBER_QUEUEGROUP
-              value: "{{ .Values.config.events.queueGroup }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_URL
+              value: "{{ .Values.config.events.nats.url }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBEPREFIX
+              value: "{{ .Values.config.events.nats.subscribePrefix }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_QUEUEGROUP
+              value: "{{ .Values.config.events.nats.queueGroup }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SOURCE
+              value: "{{ .Values.config.events.nats.source }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_CONNECTTIMEOUT
+              value: "{{ .Values.config.events.nats.connectTimeout }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SHUTDOWNTIMEOUT
+              value: "{{ .Values.config.events.nats.shutdownTimeout }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERFETCHBATCHSIZE
+              value: "{{ .Values.config.events.nats.subscriberFetchBatchSize }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERFETCHTIMEOUT
+              value: "{{ .Values.config.events.nats.subscriberFetchTimeout }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERFETCHBACKOFF
+              value: "{{ .Values.config.events.nats.subscriberFetchBackoff }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERNOACKEXPLICIT
+              value: "{{ .Values.config.events.nats.subscriberNoAckExplicit }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERNOMANUALACK
+              value: "{{ .Values.config.events.nats.subscriberNoManualAck }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERDELIVERYPOLICY
+              value: "{{ .Values.config.events.nats.subscriberDeliveryPolicy }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERSTARTSEQUENCE
+              value: "{{ .Values.config.events.nats.subscriberStartSequence }}"
           {{- with .Values.config.events.topics }}
             - name: PERMISSIONSAPI_EVENTS_TOPICS
               value: "{{ join " " . }}"
           {{- end }}
-          {{- if .Values.config.events.nats.credsSecretName }}
-            - name: PERMISSIONSAPI_EVENTS_SUBSCRIBER_NATS_CREDSFILE
-              value: "{{ .Values.config.events.nats.credsFile }}"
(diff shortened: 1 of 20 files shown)


chart/permissions-api/templates/deployment-worker.yaml
```diff
@@ -58,26 +58,38 @@ spec:
...
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERFETCHBATCHSIZE
+              value: "{{ .Values.config.events.nats.subscriberFetchBatchSize }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERFETCHTIMEOUT
+              value: "{{ .Values.config.events.nats.subscriberFetchTimeout }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERFETCHDELAY
+              value: "{{ .Values.config.events.nats.subscriberFetchDelay }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERACKEXPLICIT
+              value: "{{ .Values.config.events.nats.subscriberAckExplicit }}"
+            - name: PERMISSIONSAPI_EVENTS_NATS_SUBSCRIBERMANUALACK
+              value: "{{ .Values.config.events.nats.subscriberManualAck }}"
```
jnschaeffer: Looks like these should be updated to match the current config items in the events package.

cmd/worker.go
```diff
@@ -68,26 +71,34 @@ func worker(ctx context.Context, cfg *config.AppConfig) {
...
+
+	subscriber, err := pubsub.NewSubscriber(ctx, events, engine,
+		pubsub.WithLogger(logger),
+	)
 	if err != nil {
 		logger.Fatalw("unable to initialize subscriber", "error", err)
 	}
 
-	defer subscriber.Close()
+	topics := viper.GetStringSlice("events.topics")
```
jnschaeffer, line 86: Should we make this a config item instead of just a Viper call?

internal/config/config.go
```diff
@@ -2,18 +2,22 @@
...
+	"go.infratographer.com/x/viperx"
 
 	"go.infratographer.com/permissions-api/internal/spicedbx"
 )
 
 // EventsConfig stores the configuration for a load-balancer-api events config
 type EventsConfig struct {
-	Subscriber events.SubscriberConfig
+	events.Config `mapstructure:",squash"`
+	Topics        []string
```
jnschaeffer, line 20: Looks like we do actually have a config section for event topics; we should use this, right?

internal/pubsub/subscriber.go
```diff
@@ -103,167 +90,250 @@ func (s Subscriber) Listen() error {
...
+			elogger.Errorw("failed to process msg", "error", err)
 
-			msg.Nack()
-		} else {
-			msg.Ack()
+			if nakErr := msg.Nak(nakDelay); nakErr != nil {
+				elogger.Warnw("error occurred while naking", "error", nakErr)
+			}
+		} else if ackErr := msg.Ack(); ackErr != nil {
+			elogger.Errorw("error occurred while acking", "error", ackErr)
```
jnschaeffer, line 111: Given the nature of request/reply, do we want to nak these messages at all or to just reply with an error on the first attempt? This seems like a question around who is responsible for retrying if an otherwise good request fails.

internal/pubsub/subscriber.go
```diff
@@ -103,167 +90,250 @@ func (s Subscriber) Listen() error {
...
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
 
-func (s *Subscriber) createRelationships(ctx context.Context, msg *message.Message, resource types.Resource, additionalSubjectIDs []gidx.PrefixedID) error {
-	var relationships []types.Relationship
+func (s *Subscriber) handleCreateEvent(ctx context.Context, msg events.Request[events.AuthRelationshipRequest, events.AuthRelationshipResponse]) error {
```
jnschaeffer, line 178: In general, the use of `respondRequest` in so many places here (with the exception of actually creating the relationships) makes me think we actually should just ack any message that comes in and respond with errors if something fails. That way, the client can decide if it wants to retry or just bail. This is particularly true IMO given that both `handleCreateEvent` and `handleDeleteEvent` always return `nil`.

internal/pubsub/subscriber.go
```diff
@@ -103,167 +90,250 @@ func (s Subscriber) Listen() error {
...
+	respondRequest(ctx, elogger, msg, err)
+
+	return nil
 }
 
-func (s *Subscriber) handleDeleteEvent(ctx context.Context, msg *message.Message, changeMsg events.ChangeMessage) error {
-	resource, err := s.qe.NewResourceFromID(changeMsg.SubjectID)
-	if err != nil {
-		s.logger.Warnw("error parsing subject ID - will not reprocess", "event_type", changeMsg.EventType, "error", err.Error())
+func respondRequest(ctx context.Context, logger *zap.SugaredLogger, msg events.Request[events.AuthRelationshipRequest, events.AuthRelationshipResponse], errors ...error) {
```
jnschaeffer, line 314: Two things:
* Adding a span here would be good
* This should probably return an error so we know if we failed to respond

jnschaeffer (changes requested):

internal/pubsub/subscriber.go
jnschaeffer: Idea: Only call `Nak` if we can't actually reply to the message.

cmd/worker.go
mikemrm: yep, good call.

internal/config/config.go
mikemrm: indeed!

internal/pubsub/subscriber.go
mikemrm: good point. i think naking only on reply failure makes sense.

internal/pubsub/subscriber.go
mikemrm: correct. if an error is produced it's given back to the requestor and they should handle the retries. however as you mentioned in another review, we should retry for reply failures. So this has been left as is, and we now only return an error for reply failures.

internal/pubsub/subscriber.go
mikemrm: updated to make responseRequest return an error when attempting to reply. if the reply fails we should then nak the message and reprocess it.

internal/pubsub/subscriber.go
mikemrm: updated to start a span for pubsub.respond which sets an error if there's an issue responding to the request. this reply error is then returned resulting in a nak and will requeue and retry.

internal/pubsub/subscriber.go
```diff
@@ -103,167 +91,255 @@ func (s Subscriber) Listen() error {
...
+		if err != nil {
+			filteredErrors = append(filteredErrors, err)
+		}
 	}
 
-	return s.deleteRelationships(ctx, msg, resource)
-}
+	response := events.AuthRelationshipResponse{
+		Errors: filteredErrors,
+	}
```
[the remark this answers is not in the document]
mikemrm: Trace context is applied by the events publisher with the context provided.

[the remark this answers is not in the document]
mikemrm: https://github.com/infratographer/x/blob/1fb9cfe07da6ce3fcadf145b5d16f17f7d33be3c/events/nats_publish.go#L34-L38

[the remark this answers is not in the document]
jnschaeffer: Awesome!

go.mod
```diff
@@ -3,17 +3,15 @@ module go.infratographer.com/permissions-api
...
 	github.com/authzed/grpcutil v0.0.0-20230524151342-4caf7fd1108a
 	github.com/labstack/echo/v4 v4.11.1
-	github.com/nats-io/nats.go v1.27.1
 	github.com/pkg/errors v0.9.1
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.16.0
 	github.com/stretchr/testify v1.8.4
-	go.infratographer.com/x v0.3.4
+	go.infratographer.com/x v0.3.5-0.20230804152936-1fb9cfe07da6
```
jnschaeffer, line 14: Are we going to be on this prerelease commit for everything until v0.3.5 is cut?

jnschaeffer (approved):

PR #153, opened 2023-08-07
move chart to using a config file instead of env

mikemrm:
After updating event topics to be pulled from the config file, viper doesn't handle environment slices automatically so instead it stores a single string with all topics.

Changing to using a config file simplifies config changes and ensures the correct values are read.


jnschaeffer (approved):
Looks reasonable.

PR #170, opened 2023-09-06
Role permissions checks

jnschaeffer:
This PR adds permissions checks for role management CRUD operations in the permissions-api API.

Because permissions checks are baked into the API (i.e., they cannot and should not be disabled), this PR also adds a command, `create-role`, which can be used to bootstrap the creation of roles that can allow a subject to create other roles. The conditions for actions like `role_create` are deliberately left out of permissions-api to allow operators to define their own policies.

### DIFF
--- a/cmd/createrole.go
+++ b/cmd/createrole.go
@@ -1,0 +1,111 @@
+package cmd
+
+import (
+	"context"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"go.infratographer.com/permissions-api/internal/config"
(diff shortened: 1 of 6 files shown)


mikemrm (approved):

jnschaeffer:
One issue with this as written right now is that with the example policy, no relationships exist that convey that a role belongs to a resource. That actually seems necessary here, looking closer at the logic.

jnschaeffer:
Updated with a change that doesn't break everything.

fishnix (approved):

PR #173, opened 2023-09-07
Add permissions checks to assignments

jnschaeffer:
This commit adds permissions checks to role assignments. Rather than define new actions here, we just use the already-defined actions for role management.

### DIFF
--- a/internal/api/assignments.go
+++ b/internal/api/assignments.go
@@ -29,21 +29,40 @@ func (r *Router) assignmentCreate(c echo.Context) error {
 		return echo.NewHTTPError(http.StatusBadRequest, "error parsing request body").SetInternal(err)
 	}
 
-	subjID, err := gidx.Parse(reqBody.SubjectID)
+	assigneeID, err := gidx.Parse(reqBody.SubjectID)
 	if err != nil {
 		return echo.NewHTTPError(http.StatusBadRequest, "error parsing subject ID").SetInternal(err)
 	}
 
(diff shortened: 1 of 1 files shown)


mikemrm (approved):

internal/api/assignments.go
```diff
@@ -110,21 +148,40 @@ func (r *Router) assignmentDelete(c echo.Context) error {
...
+	if err != nil {
+		return echo.NewHTTPError(http.StatusBadRequest, "error getting resource").SetInternal(err)
+	}
+
+	resource, err := r.engine.GetRoleResource(ctx, roleResource, "")
+	if err != nil {
+		return echo.NewHTTPError(http.StatusInternalServerError, "error getting resource").SetInternal(err)
+	}
+
+	if err := r.checkActionWithResponse(ctx, subjectResource, actionRoleUpdate, resource); err != nil {
```
mikemrm, line 176:
```suggestion
	if err := r.checkActionWithResponse(ctx, subjectResource, actionRoleDelete, resource); err != nil {
```

mikemrm (changes requested):

internal/api/assignments.go
jnschaeffer: I went back and forth on this one. In this case, we're not deleting the role itself - rather, we're removing an assignment for the role. That's why I kept it as `actionRoleUpdate`.

mikemrm (approved):

PR #184, opened 2023-11-03
Update config.go to fix ignoreNoResponders

rizzza:
viper config needs some ❤️

### DIFF
--- a/pkg/permissions/config.go
+++ b/pkg/permissions/config.go
@@ -20,6 +20,6 @@ func MustViperFlags(v *viper.Viper, flags *pflag.FlagSet) {
 	flags.String("permissions-url", "", "sets the permissions url checks should be run against")
 	viperx.MustBindFlag(v, "permissions.url", flags.Lookup("permissions-url"))
 
-	flags.String("permissions-ignore-no-responders", "", "ignores no responder errors when auth relationship requests are published")
-	viperx.MustBindFlag(v, "permissions.ignoreAuthRelationshipNoResponders", flags.Lookup("permissions-ignore-no-responders"))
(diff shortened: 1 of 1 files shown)


jnschaeffer:
@rizzza Thanks for sending this along. Out of curiosity, why are we changing the type and name of the Viper value?

rizzza:
> @rizzza Thanks for sending this along. Out of curiosity, why are we changing the type and name of the Viper value?

Both setting environment (i.e. `RESOURCEPROVIDERAPI_PERMISSIONS_IGNORENORESPONDERS=true`), and straight-up using the `--permissions-ignore-no-responders` flag wasn't allowing me to set it to true in the config

pkg/permissions/config.go
```diff
@@ -20,6 +20,6 @@ func MustViperFlags(v *viper.Viper, flags *pflag.FlagSet) {
 	flags.String("permissions-url", "", "sets the permissions url checks should be run against")
 	viperx.MustBindFlag(v, "permissions.url", flags.Lookup("permissions-url"))
 
-	flags.String("permissions-ignore-no-responders", "", "ignores no responder errors when auth relationship requests are published")
-	viperx.MustBindFlag(v, "permissions.ignoreAuthRelationshipNoResponders", flags.Lookup("permissions-ignore-no-responders"))
+	flags.Bool("permissions-ignore-no-responders", false, "ignores no responder errors when auth relationship requests are published")
+	viperx.MustBindFlag(v, "permissions.ignoreNoResponders", flags.Lookup("permissions-ignore-no-responders"))
```
jnschaeffer, line 24: Alright, I see more clearly now: the config item `IgnoreNoResponders` doesn't match the flag. Looks reasonable to me.

jnschaeffer (approved):

PR #188, opened 2023-11-16
update renovate to update all otel packages in a single pr

mikemrm:
This ensures all otel packages are updates in the same pr which depend on each other.

Previously otel updates would frequently result in broken builds as many of the packages would get updated in separate renovate prs but depend on new updated packages. This should ensure they are all updated together.


jnschaeffer (approved):
Nice find!

jnschaeffer (approved):

PR #200, opened 2023-11-16
use full consistent reads

mikemrm:
This switches us to fully consistent reads of relationships which ensures after any action has been taken a subsequent read request will return the most up to date information.

This resolves an issue seen when directly after creating a role, a client may fail to assign a subject to the newly created role.


fishnix (approved):
I understand why we switched to fully consistent, but I'm wondering (not necessarily saying we should change it):
1. should this be a config option?
2. should we continue to support the token if its a config option?

mikemrm:
When we develop and implement an eventually consistent model both of these options would only make sense then. Until then, we will need to be strictly fully consistent.

PR #202, opened 2023-11-27
add named role support

mikemrm:
This adds support for giving roles names, along with tracking additional data such as the creator, created at and updated at.

To support this, a small database package has been created to handle interacting with the backend database and simplifying changes.

In addition to the database package, the http api has changed to support these new features.

- All endpoints which return a role will now return the new information if such details exist.
- `POST /api/v1/resources/:id/roles` now requires a `name` to be provided.
- `PATCH /api/v1/roles/:id` has been added with support for updating an existing roles name and actions.


chart/permissions-api/templates/job-migrate-database.yaml
```diff
@@ -0,0 +1,70 @@
+{{- if has .Values.config.crdb.migrateHook (list "pre-sync" "manual") }}
```
jnschaeffer, line 1: Do we anticipate needing or wanting to support these three kinds of migration workflows?

chart/permissions-api/templates/job-migrate-database.yaml
```diff
@@ -0,0 +1,70 @@
...
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  {{- if eq .Values.config.crdb.migrateHook "manual" }}
+  name: {{ include "common.names.name" . }}-migrate-database
+  {{- else }}
+  generateName: migrate-database-
+  annotations:
+    argocd.argoproj.io/hook: PreSync
```
jnschaeffer, line 11: In general I think it's good to not assume any user of this is using Argo specifically.

chart/permissions-api/values.yaml
```diff
@@ -46,6 +46,39 @@ config:
     # policyConfigMapName is the name of the Config Map containing the policy file configuration
     policyConfigMapName: ""
 
+  crdb:
+    # migrateHook sets when to run database migrations. one of: pre-sync, init, manual
+    #  - pre-sync: hook runs as a job before any other changes are synced.
+    #  - init: is run as an init container to the server deployment and may run multiple times if replica count is high.
+    #  - manual: a migrate-database job will be available to triggered manually
+    migrateHook: "init"
```
jnschaeffer, line 54: Same comment re: Argo. I do like the concept, though!

chart/permissions-api/values.yaml
```diff
@@ -46,6 +46,39 @@ config:
...
+    # name is the database name
+    name: ""
+    # host is the database host
+    host: ""
+    # user is the auth username to the database
+    user: ""
+    # password is the auth password to the database
+    password: ""
```
jnschaeffer, line 62: This seems like something we shouldn't even let users set manually, right?

chart/permissions-api/values.yaml
```diff
@@ -46,6 +46,39 @@ config:
...
+    # params is the connection parameters to the database
+    params: ""
+    # uri is the raw uri connection string
+    uri: ""
```
jnschaeffer, line 66: Same here re: sensitive data.

jnschaeffer (changes requested):
Dropping some early comments here regarding the Helm chart. I'll get to the code next.

chart/permissions-api/templates/job-migrate-database.yaml
mikemrm: manual is perhaps not needed. but pre-sync and init are useful.

pre-sync can't be used during the initial deployment as it will rely on configs and secrets that have to be synced first, and therefore won't be available. so you must use the init method. but after using the init method, you can switch to the pre-sync method which will ensure migrations only execute once per sync.

if having these options is not preferred we could get rid of it and just use init containers.

chart/permissions-api/values.yaml
mikemrm: I usually like to expose all available configuration options, which this is an option that can be set in the crdbx config which is why I added it here.

internal/database/testdb/testdb.go
```diff
@@ -0,0 +1,48 @@
+// Package testdb is a testing helper package which initializes a new crdb database and runs migrations
```
jnschaeffer, line 1: Be mindful that symbols like the one in this file will (probably) be included in the final binary when built, along with their dependencies. You'll either need a build tag for this or to put it in a testing file.

internal/database/transactions.go
```diff
@@ -0,0 +1,43 @@
...
+	"go.uber.org/zap"
+)
+
+// Transaction represents an in flight change being made to the database that must be committed or rolled back.
+type Transaction[T any] struct {
+	logger *zap.SugaredLogger
+	tx     *sql.Tx
+
+	Record T
+}
```
jnschaeffer, line 16: A transaction on a single record seems strange to me. What if I want a transaction on many things?

internal/query/relations.go
```diff
@@ -267,16 +269,166 @@ func (e *engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
+
+	role := newRole(roleName, actions)
 	roleRels := e.roleRelationships(role, res)
 
+	dbTx, err := e.db.CreateRoleTransaction(ctx, actor.ID, role.ID, roleName, res.ID)
+	if err != nil {
+		return types.Role{}, err
+	}
+
+	defer dbTx.Rollback() //nolint:errcheck // error is logged in function
```
jnschaeffer, line 285: Not sure how I feel about this. This means that no matter what, we'll always attempt a rollback, even if it fails.

internal/query/relations.go
```diff
@@ -267,16 +269,166 @@ func (e *engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
+		}
+	}
+
+	return add, rem
+}
+
+// UpdateRole allows for updating an existing role with a new name and new actions.
+// If new name is empty, no change is made.
+// If new actions is an empty slice, no change is made.
+func (e *engine) UpdateRole(ctx context.Context, actor, roleResource types.Resource, newName string, newActions []string) (types.Role, error) {
```
jnschaeffer, line 349: A `SELECT ... FOR UPDATE` is needed somewhere in here. Because we can't start up a single transaction for both SpiceDB and the main CRDB database, we should at least serialize operations on the same role.

internal/query/relations.go
```diff
@@ -267,16 +269,166 @@ func (e *engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
+	addActions, remActions := actionsDiff(role.Actions, newActions)
+
+	// If no changes, return existing role with no changes.
+	if role.Name == newName && len(addActions) == 0 && len(remActions) == 0 {
+		return role, nil
+	}
+
+	resourceID := role.ResourceID
+
+	// If the resource id is not found in the permissions database, then we must locate it in the spicedb database.
```
jnschaeffer, line 370: If this is the case, don't we have an error in our logic? No role should exist without a corresponding row in the main CRDB database.

internal/query/relations.go
```diff
@@ -267,16 +269,166 @@ func (e *engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
+
+	// If new name has changed, commit change to permissions database.
+	if newName != "" && role.Name != newName {
+		dbTx, dbErr = e.db.UpdateRoleTransaction(ctx, actor.ID, role.ID, newName, resourceID)
+		if dbErr != nil {
+			return types.Role{}, dbErr
+		}
+
+		defer dbTx.Rollback() //nolint:errcheck // error is logged in function
+	}
```
jnschaeffer, line 398: We always want to update the role somehow, even if it's just to change the updated timestamp.

internal/query/relations.go
```diff
@@ -631,13 +820,50 @@ func (e *engine) ListRoles(ctx context.Context, resource types.Resource) ([]type
...
+			if dbRole, ok := rolesByID[role.ID.String()]; ok {
+				role.Name = dbRole.Name
+				role.Creator = dbRole.CreatorID
+				role.CreatedAt = dbRole.CreatedAt
+				role.UpdatedAt = dbRole.UpdatedAt
+
+				out[i] = role
+			}
+		}
+	}
```
jnschaeffer, line 865: Have we determined that the time cost of getting roles from both CRDB and SpiceDB is worth the complexity of adding concurrency here?

internal/query/relations.go
```diff
@@ -717,9 +943,24 @@ func (e *engine) GetRole(ctx context.Context, roleResource types.Resource) (type
...
 		}
 
+		dbRole, err := e.db.GetRoleByID(ctx, roleResource.ID)
+		if err != nil && !errors.Is(err, database.ErrNoRoleFound) {
+			e.logger.Error("error while getting role", zap.Error(err))
+		}
+
+		if dbRole == nil {
+			dbRole = new(database.Role)
+		}
```
jnschaeffer, line 953: As before, if there's no corresponding CRDB entry for the role, that should be an error. We don't want a mismatch between the state of CRDB and the state of SpiceDB.

internal/query/relations.go
```diff
@@ -803,9 +1048,35 @@ func (e *engine) DeleteRole(ctx context.Context, roleResource types.Resource) er
...
 		if err = e.deleteRelationships(ctx, filter); err != nil {
-			return fmt.Errorf("failed to delete role action %s: %w", filter.OptionalResourceId, err)
+			err = fmt.Errorf("failed to delete role action %s: %w", filter.OptionalResourceId, err)
+
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+
+			return err
+		}
+	}
```
jnschaeffer, line 1071: What happens if I attempt to update a role with new actions while it's being deleted? Reading this code, it looks possible that I wind up with relationships in SpiceDB but no corresponding CRDB entries. This might also, ironically, require a `SELECT ... FOR UPDATE`.

jnschaeffer (changes requested):
Some thoughts.

internal/query/relations.go
mikemrm: Unless we provide a migration process that loads all existing roles from spicedb and populates crdb. Initially crdb will be empty. The expectation was that we would allow updating of an existing spicedb role with a name. The role has to exist to pass permission checks.

db/migrations.go
```diff
@@ -0,0 +1,11 @@
+// Package db provides an embedded filesystem containing all the database migrations
```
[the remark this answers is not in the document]
mikemrm: this is consistent with the other services which implement databases.

internal/query/relations.go
mikemrm: this is now handled properly as we always have a transaction open before on the database. if two changes came in at the same time, one would have to wait until the database released the row.

internal/database/roles.go
```diff
@@ -0,0 +1,254 @@
...
+		WHERE id = $1
+		`, id.String(),
+	).Scan(
+		&role.ID,
+		&role.Name,
+		&role.ResourceID,
+		&role.CreatorID,
+		&role.CreatedAt,
+		&role.UpdatedAt,
+	)
```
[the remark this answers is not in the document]
mikemrm: I took a look. I think that the way it is done currently is still easier to read currently. The example provided does allow some flexibility at the cost of readability and would still require a decent amount of manual changes if anything were to change. Such as updating the Scan argument ordering, the bind arguments etc.

internal/storage/roles.go
```diff
@@ -0,0 +1,274 @@
...
+			UPDATE roles SET updated_at = now() WHERE id = $1
+			RETURNING id, name, resource_id, creator_id, created_at, updated_at
+		`, roleID.String())
+	} else {
+		row = tx.QueryRowContext(ctx, `
+			UPDATE roles SET name = $1, updated_at = now() WHERE id = $2
+			RETURNING id, name, resource_id, creator_id, created_at, updated_at
+			`, name, roleID.String(),
+		)
+	}
```
jnschaeffer, line 216: As before, we can probably make this less complex by just accepting whatever role name we get.

jnschaeffer (changes requested):
I haven't looked at the tests yet but this looks broadly good. Some minor points below.

internal/api/types.go
```diff
@@ -5,12 +5,24 @@ import (
...
+	Actions []string `json:"actions"`
+}
+
 type roleResponse struct {
 	ID      gidx.PrefixedID `json:"id"`
+	Name    string          `json:"name"`
 	Actions []string        `json:"actions"`
+
+	ResourceID gidx.PrefixedID `json:"resource_id,omitempty"`
+	Creator    gidx.PrefixedID `json:"creator"`
```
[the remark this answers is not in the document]
mikemrm: whoops, good catch.

internal/query/mock/mock.go
```diff
@@ -43,20 +44,29 @@ func (e *Engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
+		CreatedAt: time.Now(),
+		UpdatedAt: time.Now(),
 	}
 
 	return role, nil
 }
 
+// UpdateRole returns nothing but satisfies the Engine interface.
+func (e *Engine) UpdateRole(ctx context.Context, actor, roleResource types.Resource, newName string, newActions []string) (types.Role, error) {
+	return types.Role{}, nil
```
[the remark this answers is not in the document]
mikemrm: yeah fair nuff. i'll update this to use the mock support for returning values.

internal/query/relations.go
```diff
@@ -267,19 +269,187 @@ func (e *engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
+	if err != nil {
+		return types.Role{}, err
+	}
+
+	newName = strings.TrimSpace(newName)
+
+	// If the name is the same, then clear the changed name.
+	if role.Name == newName {
+		newName = ""
+	}
```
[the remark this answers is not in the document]
mikemrm: good point. i'll update it to instead, if no name is provided (for example, when just updating actions) it will populate the name from the existing record so we can keep the query simple.

internal/query/relations.go
```diff
@@ -267,19 +269,187 @@ func (e *engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
+
+	resource, err := e.NewResourceFromID(resourceID)
+	if err != nil {
+		return types.Role{}, err
+	}
+
+	dbCtx, err := e.store.BeginContext(ctx)
+	if err != nil {
+		return types.Role{}, err
+	}
```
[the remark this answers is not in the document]
mikemrm: ok, currently the get methods don't retrieve the transaction from the contexts. so i'll update those to retrieve the transaction from the context and start the context sooner.

internal/storage/errors.go
```diff
@@ -0,0 +1,44 @@
...
+	// ErrorMissingContextTx represents an error where no context transaction was provided.
+	ErrorMissingContextTx = errors.New("no transaction provided in context")
+
+	// ErrorInvalidContextTx represents an error where the given context transaction is of the wrong type.
+	ErrorInvalidContextTx = errors.New("invalid type for transaction context")
+)
+
+func pqIsRoleAlreadyExistsError(err error) bool {
+	if pqErr, ok := err.(*pq.Error); ok {
+		return pqErr.Code.Name() == "unique_violation" && pqErr.Constraint == "roles_pkey"
```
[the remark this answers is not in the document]
mikemrm: I'm not sure of a better way to handle error detection. Since we own the database and migrations, we know the names of the indexes. We could make these more configurable or make them discovered on start but I'm not sure that's necessary at this point.

Alternatively we could just return less descriptive errors but I would like to avoid that.

internal/query/relations.go
```diff
@@ -267,19 +269,183 @@ func (e *engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
+		role.Actions = newActions
+	}
+
+	if err := e.store.CommitContext(dbCtx); err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+
+		logRollbackErr(e.logger, e.store.RollbackContext(dbCtx))
+
+		// TODO: add spicedb rollback logic.
```
senior, line 428: What happens without this?

internal/query/relations.go
```diff
@@ -267,19 +269,183 @@ func (e *engine) CreateRelationships(ctx context.Context, rels []types.Relations
...
 
+// actionsDiff determines which actions needs to be added and removed.
+// If no new actions are provided it is assumed no changes are requested.
+func actionsDiff(oldActions, newActions []string) ([]string, []string) {
+	if len(newActions) == 0 {
+		return nil, nil
+	}
+
+	old := make(map[string]bool, len(oldActions))
+	new := make(map[string]bool, len(newActions))
```
jnschaeffer, line 334: No need to change this, but if you want to save a few bytes on the heap, `struct{}` can be used here along with the `_, ok := myMap["some_key"]` pattern.

internal/query/relations.go
```diff
@@ -631,13 +834,35 @@ func (e *engine) ListRoles(ctx context.Context, resource types.Resource) ([]type
...
+	}
+
 	relationships, err := e.readRelationships(ctx, filter)
 	if err != nil {
 		return nil, err
 	}
 
 	out := relationshipsToRoles(relationships)
 
+	rolesByID := make(map[string]storage.Role, len(dbRoles))
```
jnschaeffer, line 849: Maybe a silly question, but why not just have a `map[gidx.PrefixedID]storage.Role` if we just convert every key to a string?

jnschaeffer (changes requested):
Logic looks good - mostly some notes on documentation and tests.

internal/query/relations.go
mikemrm: Thanks for that clarification, I'll get a version of this added to the code.

internal/query/relations.go
mikemrm: 🤔 that's a good question... i'm not sure why I did that... It definitely doesn't need to be converted. Thanks!

internal/storage/migrations/20231122000000_initial_schema.sql
```diff
@@ -0,0 +1,35 @@
+-- +goose Up
+-- create "roles" table
+CREATE TABLE "roles" (
+  "id" character varying NOT NULL,
+  "name" character varying(64) NOT NULL,
```
alexhall, line 5: What's the reasoning for setting this limit at the storage level? In my experience, doing this is just setting yourself up for running future database migrations to increase the limit - 64 characters is pretty low for something you intend to expose to users.

alexhall (commented):
Would it make sense to store the role actions in CRDB and anoint that as a source of truth for role definitions? If I were the one on PagerDuty rotation for this system I'd be pretty nervous about the brittleness of coordinating transactions between different databases and the risk of things getting into an inconsistent state, and having a single source of truth at least gives you a starting point for restoring the system to a known-good state when the need arises.

jnschaeffer:
> Would it make sense to store the role actions in CRDB and anoint that as a source of truth for role definitions? If I were the one on PagerDuty rotation for this system I'd be pretty nervous about the brittleness of coordinating transactions between different databases and the risk of things getting into an inconsistent state, and having a single source of truth at least gives you a starting point for restoring the system to a known-good state when the need arises.

@alexhall I missed this before, sorry. I might be wrong here, but if we have actions stored in CRDB it doesn't seem like that fixes the problems that come from coordinating transactions between both systems. If a commit fails in CRDB but we already persisted actions to SpiceDB, for example, we'll now be in an inconsistent state (i.e., SpiceDB will have a different view of actions in a role from what CRDB shows). As written, we at least keep the actions in only one place.

alexhall:
> > Would it make sense to store the role actions in CRDB and anoint that as a source of truth for role definitions? If I were the one on PagerDuty rotation for this system I'd be pretty nervous about the brittleness of coordinating transactions between different databases and the risk of things getting into an inconsistent state, and having a single source of truth at least gives you a starting point for restoring the system to a known-good state when the need arises.
> 
> @alexhall I missed this before, sorry. I might be wrong here, but if we have actions stored in CRDB it doesn't seem like that fixes the problems that come from coordinating transactions between both systems. If a commit fails in CRDB but we already persisted actions to SpiceDB, for example, we'll now be in an inconsistent state (i.e., SpiceDB will have a different view of actions in a role from what CRDB shows). As written, we at least keep the actions in only one place.

In this situation, I wouldn't be writing to SpiceDB until after the write to CRDB succeeds. If the write to SpiceDB fails, that can be retried, and possibly even reconciled in the background, but at least I'd know that I have persisted in CRDB a consistent view of how the roles and actions *should* be. In other words, treat SpiceDB as a query-optimized view used for evaluating access checks, accept that data will get out of sync between CRDB and SpiceDB, and optimize for your ability to recover from that situation. I don't know how well that aligns with your existing architecture, but that's what jumps out to me at first glance when looking at this PR.

internal/storage/migrations/20231122000000_initial_schema.sql
mikemrm: 64 characters seems to be a common limit. This field is part of a unique index and needs to have a limit to ensure performance.

jnschaeffer:
> > > Would it make sense to store the role actions in CRDB and anoint that as a source of truth for role definitions? If I were the one on PagerDuty rotation for this system I'd be pretty nervous about the brittleness of coordinating transactions between different databases and the risk of things getting into an inconsistent state, and having a single source of truth at least gives you a starting point for restoring the system to a known-good state when the need arises.
> > 
> > 
> > @alexhall I missed this before, sorry. I might be wrong here, but if we have actions stored in CRDB it doesn't seem like that fixes the problems that come from coordinating transactions between both systems. If a commit fails in CRDB but we already persisted actions to SpiceDB, for example, we'll now be in an inconsistent state (i.e., SpiceDB will have a different view of actions in a role from what CRDB shows). As written, we at least keep the actions in only one place.
> 
> In this situation, I wouldn't be writing to SpiceDB until after the write to CRDB succeeds. If the write to SpiceDB fails, that can be retried, and possibly even reconciled in the background, but at least I'd know that I have persisted in CRDB a consistent view of how the roles and actions _should_ be. In other words, treat SpiceDB as a query-optimized view used for evaluating access checks, accept that data will get out of sync between CRDB and SpiceDB, and optimize for your ability to recover from that situation. I don't know how well that aligns with your existing architecture, but that's what jumps out to me at first glance when looking at this PR.

I think this idea has merit, though perhaps for different reasons: storing the actions in CRDB lets us have more flexibility around how we present them (i.e., changing the action names, adding descriptions, etc).

That said, would it be worth putting this in a separate PR? This one is already at ~2k LOC and we probably want to give more consideration to how we define actions in general for reasons beyond DB transactions.

jnschaeffer (approved):
Looks good - at this point I think we're at least in a good enough state to end-to-end test this and see how it shakes out before cutting a release.

PR #209, opened 2024-01-14
Cache ZedTokens for resources using NATS

jnschaeffer:
Using full consistency when doing permissions checks is slow. In general, this is addressed by using ZedTokens to indicate minimum bounds on freshness when looking up cached data. Something has to keep track of those tokens, either on the client side or server-side.

This PR introduces worker caching of ZedTokens for resources on updates to relationships and updates the query engine to use those tokens when performing permissions checks. When a worker updates a relationship, it persists the ZedToken for all resources directly affected by that update to a NATS KV bucket. NATS KV writes are immediately consistent, so the new ZedToken for that resource is available to all consumers, including permissions-api API frontends. When the query engine performs a permissions check, it checks to see if a ZedToken is available for the resource.

If a ZedToken was found, that ZedToken is used along with the `at_least_as_fresh` SpiceDB API consistency strategy. If not, or if there was an error accessing NATS, the query engine falls back to the `minimize_latency` API consistency strategy. If the NATS KV bucket is configured with a TTL at least as high as the quantization interval for SpiceDB, this ensures that by the time the ZedToken is evicted from the cache, all SpiceDB frontends will be updated with data at least as fresh as the last relationship update for a resource. Clients that wish to force an update for a resource (e.g., making role changes immediately available to tenant users) can thus issue a relationship update to permissions-api and get the latest data for that resource.

This PR assumes that the KV bucket used already exists; permissions-api will not attempt to create it. This is because the intention is that the KV bucket has a TTL set to something close to the SpiceDB quantization interval, which permissions-api is not necessarily aware of.

### DIFF
--- a/internal/query/zedtokens.go
+++ b/internal/query/zedtokens.go
@@ -1,0 +1,145 @@
+package query
+
+import (
+	"context"
+	"errors"
+
+	pb "github.com/authzed/authzed-go/proto/authzed/api/v1"
+	"github.com/nats-io/nats.go"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+
+	"go.infratographer.com/permissions-api/internal/types"
+)
+
+const (
+	consistencyMinimizeLatency = "minimize_latency"
+	consistencyAtLeastAsFresh  = "at_least_as_fresh"
+)
+
+// getLatestZedToken attempts to get the latest ZedToken for the given resource ID.
+func (e *engine) getLatestZedToken(ctx context.Context, resourceID string) (string, error) {
+	_, span := e.tracer.Start(
+		ctx,
+		"getLatestZedToken",
+		trace.WithAttributes(
+			attribute.String(
+				"permissions.resource",
+				resourceID,
+			),
+		),
+	)
+
+	defer span.End()
+
+	resp, err := e.kv.Get(resourceID)
+	if err != nil {
+		// Only record this as an error if it wasn't a not found error.
+		if !errors.Is(err, nats.ErrKeyNotFound) {
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+		}
+
+		return "", err
+	}
+
+	zedToken := string(resp.Value())
+
+	return zedToken, nil
+}
+
+// upsertZedToken updates the ZedToken at the given resource ID key with the provided ZedToken.
+func (e *engine) upsertZedToken(ctx context.Context, resourceID string, zedToken string) error {
+	_, span := e.tracer.Start(
+		ctx,
+		"upsertZedToken",
+		trace.WithAttributes(
+			attribute.String(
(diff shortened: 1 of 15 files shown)


internal/query/zedtokens.go
```diff
@@ -0,0 +1,145 @@
...
+		_, err = e.kv.Update(resourceID, zedTokenBytes, resp.Revision())
+	// If we did not find a token, create it. This may fail if another client created an entry already
+	case errors.Is(getErr, nats.ErrKeyNotFound):
+		_, err = e.kv.Create(resourceID, zedTokenBytes)
+	// If something else happened, just keep moving
+	default:
+	}
+
+	// If an error happened when creating or updating the token, record it.
+	if err != nil {
```
fishnix, line 87: ill be interested to see how often we hit this condition - i expect not a lot since its for an individual relationship

internal/query/zedtokens.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// updateRelationshipZedTokens updates the NATS KV bucket for ZedTokens, setting the given ZedToken
+// as the latest point in time snapshot for every resource in the given list of relationships.
+func (e *engine) updateRelationshipZedTokens(ctx context.Context, rels []types.Relationship, zedToken string) {
+	resourceIDMap := map[string]struct{}{}
```
fishnix, line 100: why this over just a slice here since you're just looping over the keys anyways?

internal/query/zedtokens.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+}
+
+// determineConsistency produces a consistency strategy based on whether a ZedToken exists for a
+// given resource. If a ZedToken is available for the resource, at_least_as_fresh is used with the
+// retrieved ZedToken. If no such token is found, or if there is an error reaching NATS, minimize_latency
+// is used. This ensures that if NATS is not working or available for some reason, we can still make
+// permissions checks (albeit in a degraded state).
+func (e *engine) determineConsistency(ctx context.Context, resource types.Resource) (*pb.Consistency, string) {
```
fishnix, line 118: should we have a test for this?

fishnix (dismissed):
some minor stuff, this looks good 👍

mikemrm (approved):

internal/query/zedtokens.go
jnschaeffer: Relationships don't necessarily need to refer to unique resources (i.e., multiple relationships could be persisted that affect the same tenant). This just ensures that we aren't sending redundant ZedToken updates.

internal/query/zedtokens.go
jnschaeffer: Probably not often, but I can think of a few cases:
* NATS is unavailable
* Something created a bucket entry before this function did
* Something updated a bucket entry

internal/query/zedtokens.go
jnschaeffer: Sure, I can add one.

PR #217, opened 2024-02-02
permit unconfigured permission checks

mikemrm:
Easily permit permission checks when the permission library doesn't have a proper url configured.

This simplifies local development for service which implement permissions checks allowing them to skip calling out to a permissions service while developing locally by simply enabling `AllowUnconfiguredChecks`.

### DIFF
--- a/pkg/permissions/config.go
+++ b/pkg/permissions/config.go
@@ -8,11 +8,16 @@ import (
 
 // Config defines the permissions configuration structure
 type Config struct {
-	// URL is the URL checks should be executed against
+	// URL should point to a permissions-api authorization API route, such as https://example.com/api/v1/allow.
+	// If not set, all permissions checks will be denied by default. To override this behavior, set DefaultAllow
+	// to true.
 	URL string
 
 	// IgnoreNoResponders will ignore no responder errors when auth relationship requests are published.
 	IgnoreNoResponders bool
+
+	// DefaultAllow if set to true, will allow all permissions checks when URL is not set.
+	DefaultAllow bool
 }
 
 // MustViperFlags adds permissions config flags and viper bindings
@@ -22,4 +27,7 @@ func MustViperFlags(v *viper.Viper, flags *pflag.FlagSet) {
 
 	flags.Bool("permissions-ignore-no-responders", false, "ignores no responder errors when auth relationship requests are published")
 	viperx.MustBindFlag(v, "permissions.ignoreNoResponders", flags.Lookup("permissions-ignore-no-responders"))
+
+	flags.Bool("permissions-default-allow", false, "grant permission checks when url is not set")
+	viperx.MustBindFlag(v, "permissions.defaultAllow", flags.Lookup("permissions-default-allow"))
 }
--- a/pkg/permissions/permissions.go
+++ b/pkg/permissions/permissions.go
@@ -201,6 +201,10 @@ func New(config Config, options ...Option) (*Permissions, error) {
 		p.url = uri
 	}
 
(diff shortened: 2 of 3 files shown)


pkg/permissions/config.go
```diff
@@ -8,11 +8,15 @@ import (
...
-	// URL is the URL checks should be executed against
+	// URL is the URL checks should be executed against.
+	// Considered unconfigured if empty.
 	URL string
 
 	// IgnoreNoResponders will ignore no responder errors when auth relationship requests are published.
 	IgnoreNoResponders bool
+
+	// AllowUnconfiguredChecks permits permission checks when unconfigured.
+	AllowUnconfiguredChecks bool
```
jnschaeffer, line 19: This strikes me as a bit of an odd name. Maybe something like `DefaultAllow` or `DefaultAllowAll` would be better; the distinction between "configured" and "unconfigured" is confusing.

pkg/permissions/config.go
```diff
@@ -8,11 +8,15 @@ import (
 
 // Config defines the permissions configuration structure
 type Config struct {
-	// URL is the URL checks should be executed against
+	// URL is the URL checks should be executed against.
+	// Considered unconfigured if empty.
```
jnschaeffer, line 12: Something like this might convey intent more clearly:

```suggestion
	// URL should point to a permissions-api authorization API route, such as https://foo.dev/api/v1/allow.
	// If not set, all permissions checks will be denied by default. To override this behavior, set DefaultAllow
	// to true.
```

jnschaeffer (changes requested):

pkg/permissions/config.go
mikemrm: I too am not a fan of the name, however I wanted to be clear that this only takes affect when the url is not defined and therefore we're not processing standard checks.

If you still think one of these is a more appropriate name, I can update it to that.

jnschaeffer: Maybe the suggested doc comment helps?

mikemrm: indeed it does, thanks.

pkg/permissions/config.go
```diff
@@ -8,11 +8,16 @@ import (
...
-	// URL is the URL checks should be executed against
+	// URL should point to a permissions-api authorization API route, such as https://example.com/api/v1/allow.
+	// If not set, all permissions checks will be denied by default. To override this behavior, set DefaultAllow
+	// to true.
 	URL string
 
 	// IgnoreNoResponders will ignore no responder errors when auth relationship requests are published.
 	IgnoreNoResponders bool
+
+	// DefaultAllow all permissions checks will be granted when URL is not set.
```
jnschaeffer, line 19:
```suggestion
	// DefaultAllow, if set to true, will allow all permissions checks when URL is not set.
```

pkg/permissions/config.go
```diff
@@ -22,4 +27,7 @@ func MustViperFlags(v *viper.Viper, flags *pflag.FlagSet) {
 
 	flags.Bool("permissions-ignore-no-responders", false, "ignores no responder errors when auth relationship requests are published")
 	viperx.MustBindFlag(v, "permissions.ignoreNoResponders", flags.Lookup("permissions-ignore-no-responders"))
+
+	flags.Bool("default-allow", false, "grant permission checks when url is not set")
```
jnschaeffer, line 31: Should this be `permissions-default-allow`?

mikemrm: oh good catch, thank you.

jnschaeffer (approved):

PR #219, opened 2024-03-05
role-binding v2 schema and IAPL

bailinhe:
In order to separate roles from bindings in permissions-api, we need to update the generated SpiceDB schema with a definition for bindings and permissions that check both direct assignment to roles and bindings (i.e., we don’t want to break existing assignments). The scope of this task is to update permissions-api to generate a SpiceDB schema with a binding definition that associates a subject (i.e., any principal) to a role.

see [this doc](https://github.com/infratographer/permissions-api/blob/41bcf2e66f9e8de28329580fd0c987cf1ae06cd6/docs/rbac.md) for more

### DIFF
--- a/docs/rbac.md
+++ b/docs/rbac.md
@@ -1,0 +1,415 @@
+# RBAC V2
+
+## Role-Bindings
+
+The following diagram describes the most basic role-binding relationships:
+
+```mermaid
+erDiagram
+  RoleBinding }o--|| Role : role
+  RoleBinding }o--o{ Subject : subject
+  Resource ||--o{ RoleBinding : grant
+
+  RoleBinding {
+    permission read_document
+    permission write_document
+    permission etc
+  }
+
+  Role {
+    relationship read_document_rel
+    relationship write_document_rel 
+    relationship etc
+  }
+
+  Resource {
+    permission read_document
+    permission write_document
+    permission etc
+  }
+```
+
+### RBAC IAPL
+
+In the IAPL, a new `rbac` directive is introduced to define the RBAC configurations.
+
+property | yaml | type | description
+-|-|-|-
+RoleResource |`rbac.roleresource`| string | name of the resource type that represents a role.
+RoleSubjectTypes |`rbac.rolesubjecttypes`| string | a list of subject types that the relationships in a role resource will contain.
+RoleOwners |`rbac.roleowners`| []string | the list of resource types that can own a role.  These resources should be (but not limited to) organizational resources like tenant, organization, project, group, etc When a role is owned by an entity, say a group, that means this role will be available to perform role-bindings for resources that are owned by this group and its subgroups.  The RoleOwners relationship is particularly useful to limit access to custom roles.
+RoleBindingResource |`rbac.rolebindingresource`| string | name of the resource type that represents a role binding.
+RoleBindingSubjects |`rbac.rolebindingsubjects`| []string | names of the resource types that can be subjects in a role binding.
+
+For example, consider the following spicedb schema:
+
+```zed
+
+ definition user {}
+ definition client {}
+
+ definition group {
+  relation member: user | client
+ }
+
+ definition organization {
+  relation parent: organization
+  relation member: user | client | organization#member
+  relation member_role: role | organization#member_role
+  relation grant: rolebinding
+
+  permissions rolebinding_list: grant->rolebinding_list
+  permissions rolebinding_create: grant->rolebinding_create
+  permissions rolebinding_delete: grant->rolebinding_delete
+ }
+
+ definition role {
+  relation owner: organization
+  relation view_organization: user:* | client:*
+  relation rolebinding_list_rel: user:* | client:*
+  relation rolebinding_create_rel: user:* | client:*
+  relation rolebinding_delete_rel: user:* | client:*
+ }
+
+ definition role_binding {
+  relation role: role
+  relation subject: user | group#member
+  permission view_organization = subject & role->view_organization
+  permissions rolebinding_list: subject & role->rolebinding_list
+  permissions rolebinding_create: subject & role->rolebinding_create
+  permissions rolebinding_delete: subject & role->rolebinding_delete
+ }
+
+```
+
+in IAPL policy terms:
+
+- the RoleResource would be "role"
+- the RoleBindingResource would be "role_binding",
+- the RoleRelationshipSubject would be `[user, client]`.
+- the RoleBindingSubjects would be `[{name: user}, {name: group, subjectrelation: member}]`.
+
+### Roles
+
+A `Role` is a spicedb entity that contains a set of permissions as relationships,
+in the form of:
+
+  ```zed
+  role:[role_id]#[permission_a]_rel@subject:*
+  role:[role_id]#[permission_b]_rel@subject:*
+  ```
+
+`subject:*` indicates any subjects [in possession](#bindings) of the role will be granted
+those permissions.
+
+### Bindings
+
+A `RoleBinding` establishes a three-way relationship between a role,
+a resource, and the subjects, in the form of:
+
+  ```zed
+  role_binding:[rb_id]#role@role:[role_id]
+  role_binding:[rb_id]#subject@subject:[subject_id]
+  resource:[res_id]#grant@role_binding:[rb_id]
+  ```
+
+### Permission Lookups
+
+Following is an example of looking up permission `read_doc` for subject `user_1`
+on a resource.
+
+Relationships:
+
+1. create role `doc_viewer`
+
+    ```zed
+    role:doc_viewer#read_doc_rel@subject:*
+    ```
+
+2. create role-binding
+
+    ```zed
+    role_binding:rb_1#role@role:doc_viwer
+    role_binding:rb_1#subject@subject:user_1
+    resource:res_1#grant@role_binding:rb_1
+    ```
+
+Lookup:
+
+```mermaid
+flowchart TD
+  res("`Resource:
+    permission: read_doc
+  `")
+
+  grant{Grant}
+  rb("`RoleBinding: rb_1`")
+
+  role_rel{Role}
+  role(doc_viwer)
+  perm>read_doc_rel]
+
+  subj_rel{Subj}
+  subj(user_1)
+
+  res-->grant
+  grant-->rb
+
+  rb-->role_rel
+  role_rel-->role
+  role-->perm
+  perm-->permok{{ok ✅}}
+
+  rb-->subj_rel
+  subj_rel-->subj
+  subj-->subjok{{ok ✅}}
+
+  permok-->ok{{ok ✅}}
+  subjok-->ok
+```
+
+## Hierarchical Grants
+
+In a lot of permissions scenarios, permission relationships are not as clean-cut
+as the example shown above. Most of the times, role-bindings involve binding a role to
+subjects on a higher level (e.g., projects, tenant, etc.,) and the users expect
+those permissions to be propagated all the resources that it owns. Moreover,
+there are cases where, instead of a single user or client being the role-binding
+subject, IAM users expect a role can be bind to a group of subjects. The IAPL
+is modified to generate a SpiceDB schema to accommodate these more complex
+use cases.
+
+### Ownerships
+
+To accommodate inheritance of the grant relationships,
+a new property `RoleBindingV2` is added to the resource type definitions
+and a new type of `Condition` is introduced to the IAPL:
+
+```diff
+  type ResourceType struct {
+    Name          string
+    IDPrefix      string
++   RoleBindingV2 *ResourceRoleBindingV2
+    Relationships []Relationship
+  }
+
+  type Condition struct {
+    RoleBinding        *ConditionRoleBinding
++   RoleBindingV2      *ConditionRoleBindingV2
+    RelationshipAction *ConditionRelationshipAction
+  }
+```
+
+A property of `InheritPermissionsFrom []string` is defined in `ResourceRoleBindingV2`
+that allows the IAPL to generate a permission line in the SpiceDB schema that
+allows grants and roles to be inherited from its owner or parent.
+
+When `RoleBindingV2` is defined in a given `Condition`, the IAPL will look for
+the `resourcetype.RoleBindingV2.InheritPermissionsFrom` property in the resource
+type that the condition's action belongs to.
+
+For example, consider the following `ActionBinding`:
+
+```yaml
+# ...
+
+resourcetypes:
+  - name: doc
+    idprefix: doc
+    rolebindingv2:
+      inheritpermissionsfrom:
+        - owner
+  - name: tenant
+    idprefix: tenant
+    rolebindingv2:
+      inheritpermissionsfrom:
+        - parent
+
+actionbindings:
+  - actionname: read_doc
+    typename: doc
+    conditions:
+      rolebindingv2: {}
+  - actionname: read_doc
+    typename: tenant
+    conditions:
+      rolebindingv2: {}
+
+# ...
+```
+
+The IAPL will generate the following SpiceDB schema:
+
+```zed
+definition doc {
+  relation owner: tenant
+  relation grant: role_binding
+  permissions read_doc: grant->read_doc + owner->read_doc
+}
+
+definition tenant {
+  relation parent: tenant
+  relation grant: role_binding
+  permissions read_doc: grant->read_doc + parent->read_doc
+}
+```
+
+#### Ownership Example
+
+Consider the following relationships, which is based on the one defined in
+[*permissions lookups*](#permission-lookups) section:
+
+```diff
+   # create role
+   role:doc_viewer#read_doc_rel@subject:*
+   
+   # grant role doc_viwer to user_1 on tenant parent
+   role_binding:rb_1#role@role:doc_viwer
+   role_binding:rb_1#subject@subject:user_1
+   tenant:parent#grant@role_binding:rb_1
+
++  # create child tenant
++  tenant:child#parent@tenant:parent
++  # create doc resource
++  doc:doc_1#owner@tenant:child
+```
+
+Lookup:
+
+```mermaid
+flowchart TD
+  doc("`Doc:
+    permission: read_doc
+  `")
+
+  tenant_child(Tenant: Child)
+  owner_rel{Owner}
+  tenant_parent(Tenant: Parent)
+  parent_rel{Parent}
+
+  grant{Grant}
+  rb("`RoleBinding: rb_1`")
+
+  role_rel{Role}
+  role(doc_viwer)
+  perm>read_doc_rel]
+
+  subj_rel{Subj}
+  subj(user_1)
+
+  doc-->owner_rel 
+  owner_rel-->tenant_child
+  tenant_child-->parent_rel
+  parent_rel-->tenant_parent
+
+  tenant_parent-->grant
+  grant-->rb
+
+  rb-->role_rel
+  role_rel-->role
+  role-->perm
+  perm-->permok{{ok ✅}}
+
+  rb-->subj_rel
+  subj_rel-->subj
+  subj-->subjok{{ok ✅}}
+
+  permok-->ok{{ok ✅}}
+  subjok-->ok
+```
+
+### Memberships
+
+To allowing groups, or any other type of resources with memberships to be the
+role-binding subject, all that is needed is to define the membership relationship
+in the `rbac.rolebindingsubjects` directive.
+
+For example, consider the following `rbac` directive:
+
+```yaml
+rbac:
+  rolebindingsubjects:
+    - name: user
+    - name: group
+      subjectrelation: member
+```
+
+this will generate the following SpiceDB schema:
+
+```zed
+definition role_binding {
+  relation role: role
+  relation subject: user | group#member
+  permission view_doc = subject & role->view_doc
+}
+```
+
+In this case, the `role_binding` entity can have a `subject` that is either a `user`
+or a member of the `group` entity.
+
+Consider the following relationships, which is based on the one defined in
+[*permissions lookups*](#permission-lookups) section:
+
+```diff
+   # create role
+   role:doc_viewer#read_doc_rel@subject:*
+
++  # create group_1
++  group:group_1#member@user:user_1
+   
++  # instead of binding to user_1, bind to group_1
+   role_binding:rb_1#role@role:doc_viwer
+-  role_binding:rb_1#subject@subject:user_1
++  role_binding:rb_2#subject@subject:group_1#member
+
+   tenant:parent#grant@role_binding:rb_1
+```
+
+Lookup:
+
+```mermaid
+flowchart TD
+  res("`Resource:
+    permission: read_doc
+  `")
+
+  grant{Grant}
+  rb("`RoleBinding: rb_1`")
+
+  role_rel{Role}
+  role(doc_viwer)
+  perm>read_doc_rel]
+
+  subj_rel{Subj}
+  subj(group_1)
+
+  member_rel{Member}
+  member(user_1)
+
+  res-->grant
+  grant-->rb
+
+  rb-->role_rel
+  role_rel-->role
+  role-->perm
+  perm-->permok{{ok ✅}}
+
+  rb-->subj_rel
+  subj_rel-->subj
+  subj-->member_rel
+  member_rel-->member
+  member-->memberok{{ok ✅}}
+
+  permok-->ok{{ok ✅}}
+  memberok-->ok
+```
+
+## Glossary
+
+- **Subject**: The entities that permissions can be granted to, such as users, clients, or group members
+- **Role**: An entity that contains a set of permissions
+- **RoleBinding**: An entity that creates a relationship between a role and some subjects,
+  meaning that these subjects are "in possession" of the permissions defined in the role
+- **Grant**: The relationship between a role-binding and a resource, effectively creating a
+  three way relationship between a role, a resource, and the subjects
+- **Inheritance**: The ability to propagate permissions and roles from a parent resource to its children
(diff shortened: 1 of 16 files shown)


internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+		}
+
+		permsRel[i] = Relationship{
+			Relation:    perm + "_rel",
+			TargetTypes: targettypes,
+		}
+	}
+
+	// 4. create a role containing all the relationships shown above
+	var role ResourceType
```
jnschaeffer, line 443: This isn't defining a role - it's defining the role type. This should be made clear.

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+
+	if _, ok := v.rt[v.p.RBAC.RoleResource]; ok {
+		role = v.rt[v.p.RBAC.RoleResource]
+		role.Relationships = permsRel
+	} else {
+		role = ResourceType{
+			Name:          v.p.RBAC.RoleResource,
+			Relationships: permsRel,
+		}
+	}
```
jnschaeffer, line 455: Why this?

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+
+	v.rt[role.Name] = role
+
+	// 5. create a list of role owners
+	for _, owner := range v.p.RBAC.RoleOwners {
+		v.roleOwners[owner] = struct{}{}
+	}
```
jnschaeffer, line 462: Why are we creating a map based on what we already know?

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+
+	if _, ok := v.rt[v.p.RBAC.RoleBindingResource]; ok {
+		rolebinding = v.rt[v.p.RBAC.RoleBindingResource]
+		rolebinding.Relationships = []Relationship{role, subjects}
+	} else {
+		rolebinding = ResourceType{
+			Name:          v.p.RBAC.RoleBindingResource,
+			Relationships: []Relationship{role, subjects},
+		}
+	}
```
jnschaeffer, line 525: As before, why this?

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+	// 2. create relationship to subjects
+	subjects := Relationship{
+		Relation:    RolebindingSubjectRelation,
+		TargetTypes: v.p.RBAC.RoleBindingSubjects,
+	}
+
+	// 3. create a list of action-bindings representing permissions for all the
+	// actions
+	actionbindings := make([]ActionBinding, len(v.ac))
+	i := 0
```
jnschaeffer, line 483: What's the intent of this? `make([]ActionBinding, 0, len(v.ac))` and then `actionBindings = append(actionBindings, ab)` would accomplish the same thing, if a single allocation is what you're looking for.

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+
+	v.rt[role.Name] = role
+
+	// 5. create a list of role owners
+	for _, owner := range v.p.RBAC.RoleOwners {
+		v.roleOwners[owner] = struct{}{}
+	}
+}
+
+func (v *policy) expandRoleBinding() {
```
jnschaeffer, line 465: This should probably have some description of what it's meant to do.

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+			targettypes := rel.TargetTypes
+
 			for _, typeName := range rel.TargetTypeNames {
 				if u, ok := v.un[typeName]; ok {
-					typeNames = append(typeNames, u.ResourceTypeNames...)
+					if len(u.ResourceTypes) > 0 {
+						targettypes = append(targettypes, u.ResourceTypes...)
+					} else {
+						typeNames = append(typeNames, u.ResourceTypeNames...)
+					}
```
jnschaeffer, line 543: This kind of logic makes me more convinced we should be willing to break the policy language if it's worth doing here. Branching logic like this adds what looks to me to be unnecessary complexity.

internal/iapl/policy.go
```diff
@@ -315,26 +647,106 @@ func (v *policy) Schema() []types.ResourceType {
 	}
 
 	for _, b := range v.bn {
+		actionName := b.ActionName
+
+		if b.RenamePermission != "" {
+			actionName = b.RenamePermission
+		}
```
jnschaeffer, line 654: What is this?

internal/iapl/policy.go
```diff
@@ -315,26 +647,106 @@ func (v *policy) Schema() []types.ResourceType {
...
 		}
 
 		for _, c := range b.Conditions {
-			condition := types.Condition{
-				RoleBinding:        (*types.ConditionRoleBinding)(c.RoleBinding),
-				RelationshipAction: (*types.ConditionRelationshipAction)(c.RelationshipAction),
+			var conditions []types.Condition
+
+			switch {
+			case c.RoleBinding != nil:
```
jnschaeffer, line 664: Can we break these conditions out so they're easier to read? This switch statement is really long.

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+
 			resourceType.Relationships[i].TargetTypeNames = typeNames
+			resourceType.Relationships[i].TargetTypes = targettypes
+		}
+
+		// role-binding V2 configs
+		availRoleConditions := []Condition{}
+
+		// if resource type is a role-owner, add the role-relationship to the
+		// resource
```
jnschaeffer, line 561: Is this the right place to do this? It might be, but in general I think we'll have a less difficult time if permissions-api defines Role and RoleBinding resource types as early as possible and adds them to the policy.

In any case, doing that in a separate method would be better anyway.

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
 
+			for _, tn := range typeNames {
+				targettypes = append(targettypes, types.TargetType{Name: tn})
+			}
+
 			resourceType.Relationships[i].TargetTypeNames = typeNames
+			resourceType.Relationships[i].TargetTypes = targettypes
+		}
+
+		// role-binding V2 configs
```
jnschaeffer, line 557: This is not a very descriptive name.

internal/iapl/rbac.go
```diff
@@ -0,0 +1,142 @@
...
+	// RoleOwnerParentRelation is the name of the relationship that connects a role's owner to its parent.
+	RoleOwnerParentRelation = "parent"
+)
+
+// ResourceRoleBindingV2 describes the relationships that will be created
+// for a resource to support role-binding V2
+type ResourceRoleBindingV2 struct {
+	// AvailableRolesFrom is the list of resource types that can provide roles to this resource
+	AvailableRolesFrom []string
+}
```
jnschaeffer, line 29: What is the purpose of this? It's not really clear.

jnschaeffer (changes requested):
Leaving a partial review here of the policy language stuff. Moving on to the rest.

internal/iapl/rbac.go
```diff
@@ -0,0 +1,142 @@
...
+
+/*
+RBAC represents a role-based access control policy.
+  - RoleResource is the name of the resource type that represents a role.
+  - RoleRelationshipSubject is the name of the relationship that connects a role to a subject.
+  - RoleOwners is the names of the resource types that can own a role.
+  - RoleBindingResource is the name of the resource type that represents a role binding.
+  - RoleBindingSubjects is the names of the resource types that can be subjects in a role binding.
+  - RolebindingPermissionsPrefix generates the permissions sets to manage role bindings,
+  - GrantRelationship is the name of the relationship that connects a role binding to a resource.
```
jnschaeffer, line 39: A few points here:
* I'm not sure `RoleResource`, `RoleBindingResource`, `RoleBindingPermissionsPrefix`, or `GrantRelationship` need to be configurable. What's gained by doing that?
* `RoleRelationshipSubject` doesn't feel like a descriptive name for what we're talking about here. What is a role relationship?
* What does it mean to own a role?

internal/iapl/rbac.go
```diff
@@ -0,0 +1,142 @@
...
+	return RBAC{
+		RoleResource:                 "rolev2",
+		RoleRelationshipSubjects:     []string{"user", "client"},
+		RoleOwners:                   []string{"tenant"},
+		RoleBindingResource:          "role_binding",
+		RoleBindingSubjects:          []types.TargetType{{Name: "user"}, {Name: "client"}, {Name: "group", SubjectRelation: "member"}},
+		RoleBindingPermissionsPrefix: "rolebinding",
+		GrantRelationship:            "grant",
+	}
+}
```
jnschaeffer, line 142: IMO we should get rid of the default policy and default RBAC stuff altogether.

internal/query/relations.go
```diff
@@ -760,7 +760,8 @@ func (e *engine) relationshipsToNonRoles(rels []*pb.Relationship) ([]types.Relat
 	var out []types.Relationship
 
 	for _, rel := range rels {
-		if rel.Subject.Object.ObjectType == e.namespace+"/role" {
+		// skip relationships for v1 roles, and wildcard relationships for v2 roles
+		if rel.Subject.Object.ObjectType == e.namespace+"/role" || rel.Subject.Object.ObjectId == "*" {
```
jnschaeffer, line 764: Magic strings here feel like they're going to introduce problems. We have a configurable role resource but then have `/role` hard-coded here.

jnschaeffer (commented):
More thoughts. Overall, the direction of this is good, but it looks like we're adding more customizability than may be warranted and the code could stand to be broken down into smaller, more easily consumed units than it currently is.

internal/iapl/policy.go
```diff
@@ -15,48 +15,67 @@ type PolicyDocument struct {
...
 // Union represents a named union of multiple concrete resource types.
 type Union struct {
 	Name              string
 	ResourceTypeNames []string
+	ResourceTypes     []types.TargetType
 }
 
 // Action represents an action that can be taken in an authorization policy.
+// AdditionalPermissions can be defined to allow additional permissions to be
+// created in addition to the permission with the same name as the action.
```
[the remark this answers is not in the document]
bailinhe: This is something I needed before, but not anymore. removed.

internal/iapl/policy.go
```diff
@@ -15,48 +15,67 @@ type PolicyDocument struct {
...
 	IDPrefix      string
+	RoleBindingV2 *ResourceRoleBindingV2
 	Relationships []Relationship
 }
 
 // Relationship represents a named relation between two resources.
 type Relationship struct {
 	Relation        string
 	TargetTypeNames []string
+	TargetTypes     []types.TargetType
```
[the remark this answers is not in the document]
bailinhe: this is to support defining relationships with subject relations, like `...owner@group#owner` or `...memebr:group@member`. Then template also needs this information to render the relationship properly.

internal/iapl/policy.go
```diff
@@ -15,48 +15,67 @@ type PolicyDocument struct {
...
+// created in addition to the permission with the same name as the action.
 type Action struct {
 	Name string
 }
 
 // ActionBinding represents a binding of an action to a resource type or union.
+// RenamePermission is the name of the permission that should be used instead of
+// the action name when checking for permission to perform the action, this
+// allows the permission to be renamed to avoid conflicts with relationships
+// with the same name.
```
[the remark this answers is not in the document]
bailinhe: the use case has gone, two re-factors ago.

internal/iapl/policy.go
```diff
@@ -93,16 +117,39 @@ func NewPolicy(p PolicyDocument) Policy {
...
+	// role-binding, delete role-binding)
+	rbActions := []string{"create", "list", "get", "update", "delete"}
+	rolebindingActions := make([]string, len(rbActions))
+	ac := map[string]Action{}
+
 	for _, a := range p.Actions {
 		ac[a.Name] = a
 	}
 
+	if p.RBAC != nil {
```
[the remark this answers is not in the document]
bailinhe: This is mainly for backward compatibility. This allows a policy to have RBAC V2 completely disabled.

internal/query/relations.go
bailinhe: only v2 roles are configurable, v1 role will remain hard coded as `role`

internal/iapl/rbac.go
bailinhe: - I do think `RoleResource` and `RoleBindingResource` should be configurable to avoid conflicted resource names, the rest I agree there's no need for the configurabilities, so I have them removed.
- ~`RoleRelationshipSubject`~ `RoleSubjectTypes`
- RoleOwners is the list of resource types that can own a role.
These resources should be (but not limited to) organizational resources
like tenant, organization, project, group, etc
When a role is owned by an entity, say a group, that means this role
will be available to perform role-bindings for resources that are owned
by this group and its subgroups.
The RoleOwners relationship is particularly useful to limit access to
custom roles. (this description is also added to the comment)

internal/iapl/policy.go
```diff
@@ -229,14 +306,79 @@ func (v *policy) validateActionBindings() error {
 	return nil
 }
 
+func (v *policy) validateRoles() error {
+	if v.p.RBAC == nil {
+		return nil
+	}
+
+	for _, roleOwnerName := range v.p.RBAC.RoleOwners {
+		owner, ok := v.rt[roleOwnerName]
```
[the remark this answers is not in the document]
bailinhe: moved this to [`struct RBAC`](https://github.com/infratographer/permissions-api/pull/219/commits/41bcf2e66f9e8de28329580fd0c987cf1ae06cd6#diff-36f723da3c50c13c2353a47bf4a16fb61e82af714eb14f32ec4403d440e67110R204)

internal/iapl/policy.go
```diff
@@ -93,16 +117,39 @@ func NewPolicy(p PolicyDocument) Policy {
...
+		ac:                 ac,
+		p:                  p,
+		permissions:        []string{},
+		rolebindingActions: rolebindingActions,
+		roleOwners:         map[string]struct{}{},
+	}
+
+	if p.RBAC != nil {
+		out.expandRole()
+		out.expandRoleBinding()
```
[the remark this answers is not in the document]
bailinhe: [this](https://github.com/infratographer/permissions-api/pull/219/commits/41bcf2e66f9e8de28329580fd0c987cf1ae06cd6#diff-05196818998625f7c1b206f8a36448bb71c85755cb020c1b1e91d8cc2dc936aaR107-R130) looks better?

internal/iapl/policy.go
```diff
@@ -229,14 +306,79 @@ func (v *policy) validateActionBindings() error {
...
+		// check if role owner exists
+		if !ok {
+			return fmt.Errorf("%w: role owner %s does not exists", ErrorUnknownType, roleOwnerName)
+		}
+
+		relationshipExists := false
+		targetOk := false
+
+		// role owner must have `member_role` relationship to `role` resource
+	rel_loop:
```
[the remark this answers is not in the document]
bailinhe: maybe [a function](https://github.com/infratographer/permissions-api/pull/219/files#diff-05196818998625f7c1b206f8a36448bb71c85755cb020c1b1e91d8cc2dc936aaR308-R337) would be better suited for this

internal/iapl/policy.go
```diff
@@ -257,20 +399,205 @@ func (v *policy) expandActionBindings() {
...
+	// 1. create a list of all permissions
+	perms := make(map[string]struct{})
+
+	for _, action := range v.ac {
+		perms[action.Name] = struct{}{}
+	}
+
+	for perm := range perms {
+		v.permissions = append(v.permissions, perm)
+	}
```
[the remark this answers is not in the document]
bailinhe: I feel like I needed this before for some de-duplication purpose and then they are no longer necessary. removed.

internal/iapl/rbac.go
```diff
@@ -0,0 +1,214 @@
...
+type ResourceRoleBindingV2 struct {
+	// AvailableRolesFrom is the list of resource types that can provide roles to this resource
+	// Note that not all roles are available to all resources. This relationship is used to
+	// determine which roles are available to a resource.
+	// Before creating a role binding for a resource, one should check whether or
+	// not the role is available for the resource.
+	//
+	// Also see the RoleOwners field in the RBAC struct
+	AvailableRolesFrom []string
+}
```
jnschaeffer, line 56: Three things:

* `ResourceRoleBindingV2` doesn't really convey what this is for, but I'm not sure what would
* `AvailableRolesFrom` as a name feels confusing, but I'm not sure what's better. I'll think about it
* The comment for `AvailableRolesFrom` is not correct here as far as I can tell. It's a list of relations, right?

internal/iapl/policy.go
```diff
@@ -43,20 +47,29 @@ type Action struct {
 
 // ActionBinding represents a binding of an action to a resource type or union.
 type ActionBinding struct {
-	ActionName string
-	TypeName   string
-	Conditions []Condition
+	ActionName    string
+	TypeName      string
+	Conditions    []Condition
+	ConditionSets []types.ConditionSet
```
jnschaeffer, line 53: What is the relationship between `Conditions` and `ConditionSets`?

internal/iapl/policy.go
```diff
@@ -43,20 +47,29 @@ type Action struct {
...
 }
 
 // ConditionRoleBinding represents a condition where a role binding is necessary to perform an action.
 type ConditionRoleBinding struct{}
 
+// ConditionRoleBindingV2 represents a condition where a role binding is necessary to perform an action.
+// Using this condition type in the policy will instruct the policy engine to
+// create all the necessary relationships in the schema to support RBAC V2
+type ConditionRoleBindingV2 struct {
+	InheritGrantsFrom []string
```
jnschaeffer, line 70: Is there a reason to have this separate from `AvailableRolesFrom`? I can't really see a scenario where I would want to have roles available from one place, but inherit grants from another.

internal/iapl/policy.go
```diff
@@ -138,19 +163,48 @@ func (v *policy) validateUnions() error {
...
-	for _, resourceType := range v.p.ResourceTypes {
+	findRelationship := func(rels []Relationship, name string) bool {
+		for _, rel := range rels {
+			if rel.Relation == name {
+				return true
+			}
+		}
+
+		return false
+	}
```
jnschaeffer, line 186: In general, if we have functions inside other functions, we should move them out unless there's a good reason (such as callback logic or some kind of closure over the environment). This doesn't seem like either of those, so it should probably be safe to move out.

docs/rbac.md
```diff
@@ -0,0 +1,387 @@
+# RBAC V2
```
jnschaeffer, line 1: Can we get a jargon table in here for things like inheritance, grants, and so on?

internal/iapl/policy.go
```diff
@@ -229,14 +287,86 @@ func (v *policy) validateActionBindings() error {
...
+	for _, roleOwnerName := range v.p.RBAC.RoleOwners {
+		owner, ok := v.rt[roleOwnerName]
+
+		// check if role owner exists
+		if !ok {
+			return fmt.Errorf("%w: role owner %s does not exist", ErrorUnknownType, roleOwnerName)
+		}
+
+		// role owner must have `member_role` relationship to `role` resource
+		ensureMemberRole := func() bool {
```
jnschaeffer, line 308: Can we make this a separate function?

jnschaeffer (commented):
More thoughts.

internal/iapl/policy.go
```diff
@@ -229,14 +287,86 @@ func (v *policy) validateActionBindings() error {
...
+					}
+				}
+
+				if !targetOk {
+					return false
+				}
+			}
+
+			return relationshipExists && targetOk
+		}
```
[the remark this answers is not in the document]
bailinhe: The `expandRBACV2Relationships` injects role owner relationships to all the role owner resources.

This function will then throws an error when either:
- the `roleOwner` defined in `rbac.roleowners` are missing from resource type definitions
- or the `roleOwner` is defined but there's no `member_role` relationship in the resource type

internal/iapl/policy.go
```diff
@@ -257,20 +387,197 @@ func (v *policy) expandActionBindings() {
...
 }
 
+// createV2RoleResourceType creates a v2 role resource type contains a list of relationships
+// representing all the actions, as well as relationships and permissions for
+// the management of the roles themselves.
+func (v *policy) createV2RoleResourceType() {
+	role, ok := v.rt[v.p.RBAC.RoleResource]
+	if !ok {
+		panic("v2 role is specified but role resource type is not defined")
+	}
```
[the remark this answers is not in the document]
bailinhe: the original thought of this is to make sure the policy developer have the `role` or `rolebinding` resource defined as a resource type, just to make sure the `idprefix` is presented. I now added the `idprefix` to be part of the `rbac` and got rid of this logics

internal/iapl/policy.go
```diff
@@ -257,20 +361,183 @@ func (v *policy) expandActionBindings() {
...
+					{Name: v.p.RBAC.RoleResource.Name},
+				},
+			}
+
+			resourceType.Relationships = append(resourceType.Relationships, memberRoleRelation)
+
+			// i.e. avail_role = member_role
+			availableRoles = append(availableRoles, Condition{
+				RelationshipAction: &ConditionRelationshipAction{
+					Relation: RoleOwnerMemberRoleRelation,
```
jnschaeffer, line 495: Is this meant to say that a role is available if a relationship exists from the owner to the role?

internal/iapl/rbac.go
```diff
@@ -0,0 +1,222 @@
...
+
+const (
+	// RoleOwnerRelation is the name of the relationship that connects a role to its owner.
+	RoleOwnerRelation = "owner"
+	// RoleOwnerMemberRoleRelation is the name of the relationship that connects a resource
+	// to a role that it owns
+	RoleOwnerMemberRoleRelation = "member_role"
+	// AvailableRoleRelation is the name of the relationship that list all roles
+	// that are available for a given resource
+	AvailableRoleRelation = "avail_role"
```
jnschaeffer, line 15: This seems like it could be named a little better, since it's actually used as an action everywhere, not as a relation. Maybe this would be better? Not sure.

```suggestion
	// UseRoleAction is the name of the action that allows the use of a role
	// for listing, binding, etc in the context of a given resource
	UseRoleAction = "iam_use_role"
```

internal/iapl/policy.go
```diff
@@ -257,20 +361,183 @@ func (v *policy) expandActionBindings() {
...
+							RelationshipAction: &types.ConditionRelationshipAction{
+								Relation:   RolebindingRoleRelation,
+								ActionName: actionName + PermissionRelationSuffix,
+							},
+						},
+					},
+				},
+				{
+					Conditions: []types.Condition{
+						{RelationshipAction: &types.ConditionRelationshipAction{Relation: RolebindingSubjectRelation}},
```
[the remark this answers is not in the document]
bailinhe: yes, that is indeed the case here.

internal/iapl/policy.go
bailinhe: yes. combined with the following code it constructs a concept of
"the roles that are available to you, as a resource, are the roles that are available to your owner or parent, plus the roles that you own (if you're a role owner)"

internal/query/relations.go
bailinhe: `git blame` 😈

internal/iapl/rbac.go
```diff
@@ -0,0 +1,222 @@
...
+```
+in IAPL policy terms:
+- the RoleResource would be "{name: role, idprefix: someprefix}"
+- the RoleBindingResource would be "{name: rolebinding, idprefix: someprefix}",
+- the RoleRelationshipSubject would be `[user, client]`.
+- the RoleBindingSubjects would be `[{name: user}, {name: group, subjectrelation: member}]`.
+*/
+type RBAC struct {
+	// RoleResource is the name of the resource type that represents a role.
+	RoleResource RBACResourceDefinition
```
jnschaeffer, line 109: Is `RBACResourceDefinition` here so we can create the resource type?

internal/spicedbx/schema.go
```diff
@@ -8,26 +8,55 @@ import (
...
-{{- end }}
-
-{{- range .Actions }}
-{{- $actionName := .Name }}
-    permission {{ $actionName }} = {{ range $index, $cond := .Conditions -}}{{ if $index }} + {{end}}{{ if $cond.RoleBinding }}{{ $actionName }}_rel{{ end }}{{ if $cond.RelationshipAction }}{{ $cond.RelationshipAction.Relation}}->{{ $cond.RelationshipAction.ActionName }}{{ end }}{{- end }}
+    permission {{ .Name }} = {{ if gt (len .Conditions) 0 }}
+			{{- template "renderCondition" . }}
+		{{- else if gt (len .ConditionSets) 0 }}
+			{{- template "renderConditionSet" . }}
+		{{- end}}
```
jnschaeffer, line 56: Defining one or the other seems unnecessary (and maybe confusing). Maybe we can converge on just one?

jnschaeffer: Update: Looks like we can move that to a separate PR.

internal/iapl/rbac.go
bailinhe: Yes. I changed that to be not private because I realized at some point we need to create a policy for tests

jnschaeffer (approved):

jnschaeffer (approved):
Looks good enough to build on. Work on things like moving `ConditionSets` into conditions and fixing the generated Mermaid graph will come later.

PR #221, opened 2024-03-07
wrap errors for easier handling in external services

mikemrm:
Services wanting to check if the error returned was an auth error or some other error could get a number of auth related errors.

This wraps the permissions and auth errors into easily comparable root errors for easier checks.

Additionally this corrects the incorrectly returned echo errors which were missing for PermissionDenied errors which resulted in 500 errors.

### DIFF
--- a/pkg/permissions/errors.go
+++ b/pkg/permissions/errors.go
@@ -1,20 +1,34 @@
 package permissions
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+
+	"github.com/labstack/echo/v4"
+)
 
 var (
+	// Error is the root error for all permissions related errors.
+	Error = errors.New("permissions error")
+
+	// AuthError is the root error all auth related errors stem from.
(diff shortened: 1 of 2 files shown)


pkg/permissions/errors.go
```diff
@@ -1,20 +1,34 @@
...
 var (
+	// Error is the root error for all permissions related errors.
+	Error = errors.New("permissions error")
+
+	// AuthError is the root error all auth related errors stem from.
+	AuthError = fmt.Errorf("%w: auth", Error) //nolint:revive,stylecheck // not returned directly, but used as a root error.
+
 	// ErrNoAuthToken is the error returned when there is no auth token provided for the API request
-	ErrNoAuthToken = errors.New("no auth token provided for client")
+	ErrNoAuthToken = echo.ErrBadRequest.WithInternal(fmt.Errorf("%w: no auth token provided for client", AuthError))
```
jnschaeffer, line 18: Wrapping these makes sense, but is this the right place for it? The permissions client and Echo middleware aren't really the same thing.

jnschaeffer (approved):

PR #222, opened 2024-03-08
Increase role http error and test quality

mikemrm:
This begins adding better request error handling.

Adds action validation checking against the policy before spending cycles hitting spicedb. This allows us to return better errors, 400's instead of 500's.

Also handles client disconnects gracefully. If the client disconnects, context canceled errors are captured and status is changed to 422's instead of an incorrect 500.

This also adds http tests for roles which resulted in the testauth package to simplify tests as auth is required in the http endpoints. This package will probably at some point be moved into the infratographer/x repo. But for now, while we develop the requirements it will be easier to develop it within the needed repo.


internal/api/router_test.go
```diff
@@ -0,0 +1,134 @@
...
+		select {
+		case <-c.Request().Context().Done():
+			return c.Request().Context().Err()
+		case <-time.After(time.Second):
+		}
+
+		switch errType {
+		case "":
+		case "echo":
+			return echo.ErrTeapot
```
jacobsee, line 37: 🫖

jacobsee (approved):

PR #231, opened 2024-04-04
Multiple Policy Files

jacobsee:
Update policy app config & config loader to support multiple policy files.
Files are merged and the merged policy document is validated after all files are loaded.
Added a check to disallow multiple bindings of the same action to the same type.


bailinhe (dismissed):
LGTM 🚀

internal/iapl/policy.go
```diff
@@ -127,6 +127,35 @@ func NewPolicyFromFile(filePath string) (Policy, error) {
...
+			return nil, err
+		}
+
+		mergedPolicy.ActionBindings = append(mergedPolicy.ActionBindings, filePolicy.ActionBindings...)
+
+		mergedPolicy.Actions = append(mergedPolicy.Actions, filePolicy.Actions...)
+
+		mergedPolicy.ResourceTypes = append(mergedPolicy.ResourceTypes, filePolicy.ResourceTypes...)
+
+		mergedPolicy.Unions = append(mergedPolicy.Unions, filePolicy.Unions...)
```
mikemrm, line 153: Something that worries me, is the policy document getting extended and this function getting forgotten about.

Possibly having a method on the policy document that accepts another document and merges them so the logic is kept within the struct methods and a little closer to home feels safer? But I'm not sure how realistic my concern is. But thought I'd mention it.

mikemrm (dismissed):
Looks good. I think there might be one potential change. But I'm not sure it a concern we might really have.

jnschaeffer (changes requested):
I will be bad cop here: Let's document the merge logic in a Markdown file in the docs directory.

jnschaeffer (dismissed):

mikemrm (changes requested):
I believe the only thing missing now is the values.yaml doesn't have a reference which is helpful for someone trying to figure out the configuration the the helm values.

mikemrm (dismissed):

internal/iapl/policy.go
```diff
@@ -127,6 +140,47 @@ func NewPolicyFromFile(filePath string) (Policy, error) {
...
+		return nil, err
+	}
+
+	filePaths := make([]string, 0, len(files))
+
+	for _, file := range files {
+		if !file.IsDir() {
+			filePaths = append(filePaths, directoryPath+"/"+file.Name())
+		}
+	}
```
mikemrm, line 179: Since we only accept yaml files, we might want to exclude non `.yml` or `.yaml` extensions.

mikemrm (approved):

PR #244, opened 2024-04-12
support directory walking and multiple documents per file

mikemrm:
This extends the directory support to include recursive directories for further organization.

Additionally, this adds support for multiple yaml documents in a single file which allows for simple organization without cluttering folders unnecessarily.


cmd/schema.go
```diff
@@ -74,17 +74,28 @@ func writeSchema(_ context.Context, dryRun bool, cfg *config.AppConfig) {
...
+				}
+
+				return nil
+			})
+
+			if err != nil {
+				logger.Fatalw("failed to read policy files from directory", "error", err)
 			}
 
 			outputPolicyMermaid(filePaths, viper.GetBool("mermaid-markdown"))
```
jnschaeffer, line 101: It looks like this code path ultimately leads to creating a policy anyway. Can we just use `NewPolicyFromDirectory` here rather than duplicate logic?

jnschaeffer (changes requested):

cmd/schema.go
mikemrm: We cannot, as `NewPolicyFromDirectory` returns a `Policy` which is a compiled policy, where as the mermaid document works off the `PolicyDocument` structure.

Once converted to a `Policy` we lose some helpful details for generating the mermaid document.

However we could create just make `NewPolicyDocumentFromDirectory` which contains the logic that both `NewPolicyFromDirectory` and the mermaid code could use.

mikemrm: Split this out into `LoadPolicyDocumentFromDirectory` which now both can use.

PR #249, opened 2024-04-22
RBAC v2 role bindings

bailinhe:
continue work on #245 

The API should be updated to use them in place of direct role assignments. The scope of this task is to make the following API changes:

Roles

Create appropriate action relationships from the roles to subject:* type relationships rather than relationships from other resources to the roles

Assignments

Create a binding when creating an assignment rather than creating a direct membership to the role

List both bindings and direct memberships when listing assignments (and show an ID for bindings)

Delete both bindings and direct memberships when deleting assignments

Check permissions based on binding_{create,get,update,delete} actions rather than role-specific actions

See [this OpenAPI specs](https://github.com/infratographer/permissions-api/blob/cb0934db964f2d3638810ed608351267c4d7659e/openapi-v2.yaml) for full list of APIs

### DIFF
--- a/internal/api/rolebindings.go
+++ b/internal/api/rolebindings.go
@@ -14,18 +14,6 @@ import (
 	"go.infratographer.com/permissions-api/internal/types"
 )
 
-func resourceToSubject(subjects []types.RoleBindingSubject) []roleBindingSubject {
-	resp := make([]roleBindingSubject, len(subjects))
-	for i, subj := range subjects {
-		resp[i] = roleBindingSubject{
-			ID:   subj.SubjectResource.ID,
-			Type: subj.SubjectResource.Type,
-		}
-	}
-
-	return resp
-}
-
 func (r *Router) roleBindingCreate(c echo.Context) error {
 	resourceIDStr := c.Param("id")
 
@@ -72,17 +60,16 @@ func (r *Router) roleBindingCreate(c echo.Context) error {
 		return r.errorResponse("error creating role resource", err)
 	}
 
-	subjects := make([]types.RoleBindingSubject, len(body.Subjects))
+	subjects := make([]types.RoleBindingSubject, len(body.SubjectIDs))
 
-	for i, s := range body.Subjects {
-		subj, err := r.engine.NewResourceFromID(s.ID)
+	for i, sid := range body.SubjectIDs {
+		subj, err := r.engine.NewResourceFromID(sid)
 		if err != nil {
 			return r.errorResponse("error creating subject resource", err)
 		}
 
 		subjects[i] = types.RoleBindingSubject{
 			SubjectResource: subj,
-			Condition:       nil,
 		}
 	}
 
@@ -96,11 +83,8 @@ func (r *Router) roleBindingCreate(c echo.Context) error {
 		roleBindingResponse{
 			ID:         rb.ID,
 			ResourceID: rb.ResourceID,
-			Subjects:   resourceToSubject(rb.Subjects),
-			Role: roleBindingResponseRole{
-				ID:   rb.Role.ID,
-				Name: rb.Role.Name,
-			},
+			SubjectIDs: rb.SubjectIDs,
+			RoleID:     rb.RoleID,
 
 			CreatedBy: rb.CreatedBy,
 			UpdatedBy: rb.UpdatedBy,
@@ -151,11 +135,8 @@ func (r *Router) roleBindingsList(c echo.Context) error {
 		resp.Data[i] = roleBindingResponse{
 			ID:         rb.ID,
 			ResourceID: rb.ResourceID,
-			Subjects:   resourceToSubject(rb.Subjects),
-			Role: roleBindingResponseRole{
-				ID:   rb.Role.ID,
-				Name: rb.Role.Name,
-			},
+			SubjectIDs: rb.SubjectIDs,
+			RoleID:     rb.RoleID,
 
 			CreatedBy: rb.CreatedBy,
 			UpdatedBy: rb.UpdatedBy,
@@ -167,7 +148,7 @@ func (r *Router) roleBindingsList(c echo.Context) error {
 	return c.JSON(http.StatusOK, resp)
 }
 
-func (r *Router) roleBindingsDelete(c echo.Context) error {
+func (r *Router) roleBindingDelete(c echo.Context) error {
 	rbID := c.Param("rb_id")
 
 	ctx, span := tracer.Start(
@@ -259,11 +240,8 @@ func (r *Router) roleBindingGet(c echo.Context) error {
 		roleBindingResponse{
 			ID:         rb.ID,
 			ResourceID: rb.ResourceID,
-			Subjects:   resourceToSubject(rb.Subjects),
-			Role: roleBindingResponseRole{
-				ID:   rb.Role.ID,
-				Name: rb.Role.Name,
-			},
+			SubjectIDs: rb.SubjectIDs,
+			RoleID:     rb.RoleID,
 
 			CreatedBy: rb.CreatedBy,
 			UpdatedBy: rb.UpdatedBy,
@@ -318,17 +296,16 @@ func (r *Router) roleBindingUpdate(c echo.Context) error {
 		return r.errorResponse(err.Error(), ErrParsingRequestBody)
 	}
 
-	subjects := make([]types.RoleBindingSubject, len(body.Subjects))
+	subjects := make([]types.RoleBindingSubject, len(body.SubjectIDs))
 
-	for i, s := range body.Subjects {
-		subj, err := r.engine.NewResourceFromID(s.ID)
+	for i, sid := range body.SubjectIDs {
+		subj, err := r.engine.NewResourceFromID(sid)
 		if err != nil {
 			return r.errorResponse("error creating subject resource", err)
 		}
 
 		subjects[i] = types.RoleBindingSubject{
 			SubjectResource: subj,
-			Condition:       nil,
 		}
 	}
 
@@ -342,12 +319,8 @@ func (r *Router) roleBindingUpdate(c echo.Context) error {
 		roleBindingResponse{
 			ID:         rb.ID,
 			ResourceID: rb.ResourceID,
-			Subjects:   resourceToSubject(rb.Subjects),
-
-			Role: roleBindingResponseRole{
-				ID:   rb.Role.ID,
-				Name: rb.Role.Name,
-			},
+			SubjectIDs: rb.SubjectIDs,
+			RoleID:     rb.RoleID,
 
 			CreatedBy: rb.CreatedBy,
 			UpdatedBy: rb.UpdatedBy,
--- a/internal/api/rolebindings.go
+++ b/internal/api/rolebindings.go
@@ -1,0 +1,366 @@
+package api
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/labstack/echo/v4"
+	"go.infratographer.com/x/gidx"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+
+	"go.infratographer.com/permissions-api/internal/iapl"
+	"go.infratographer.com/permissions-api/internal/types"
+)
+
+func resourceToSubject(subjects []types.RoleBindingSubject) []roleBindingSubject {
+	resp := make([]roleBindingSubject, len(subjects))
+	for i, subj := range subjects {
+		resp[i] = roleBindingSubject{
+			ID:        subj.SubjectResource.ID,
+			Type:      subj.SubjectResource.Type,
+			Condition: nil,
+		}
+	}
+
+	return resp
+}
+
+func (r *Router) roleBindingCreate(c echo.Context) error {
+	resourceIDStr := c.Param("id")
+
+	ctx, span := tracer.Start(
+		c.Request().Context(), "api.roleBindingCreate",
+		trace.WithAttributes(attribute.String("id", resourceIDStr)),
+	)
+	defer span.End()
+
+	resourceID, err := gidx.Parse(resourceIDStr)
+	if err != nil {
+		return r.errorResponse("error parsing resource ID", fmt.Errorf("%w: %s", ErrInvalidID, err.Error()))
+	}
+
+	var body roleBindingRequest
+
+	err = c.Bind(&body)
+	if err != nil {
+		return r.errorResponse(err.Error(), ErrParsingRequestBody)
+	}
+
+	resource, err := r.engine.NewResourceFromID(resourceID)
+	if err != nil {
+		return r.errorResponse("error creating resource", err)
+	}
+
+	subjectResource, err := r.currentSubject(c)
+	if err != nil {
+		return err
+	}
+
+	// permissions on role binding actions, similar to roles v1, are granted on the resources
+	if err := r.checkActionWithResponse(ctx, subjectResource, string(iapl.RoleBindingActionCreate), resource); err != nil {
+		return err
+	}
+
+	roleID, err := gidx.Parse(body.RoleID)
+	if err != nil {
+		return r.errorResponse("error parsing role ID", fmt.Errorf("%w: %s", ErrInvalidID, err.Error()))
+	}
+
+	roleResource, err := r.engine.NewResourceFromID(roleID)
+	if err != nil {
+		return r.errorResponse("error creating role resource", err)
+	}
+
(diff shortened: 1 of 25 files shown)


internal/api/rolebindings.go
```diff
@@ -0,0 +1,366 @@
...
+				ID:   rb.Role.ID,
+				Name: rb.Role.Name,
+			},
+		},
+	)
+}
+
+func (r *Router) roleBindingsList(c echo.Context) error {
+	resourceIDStr := c.Param("id")
+	roleIDStr := c.QueryParam("role_id")
```
jnschaeffer, line 109: Does enabling filtering by a specific role ID on the resource give us anything yet? Seems like a feature we can add in later.

internal/api/router.go
```diff
@@ -86,6 +86,12 @@ func (r *Router) Routes(rg *echo.Group) {
 		v2.PATCH("/roles/:role_id", r.roleV2Update)
 		v2.DELETE("/roles/:id", r.roleV2Delete)
 
+		v2.GET("/resources/:id/role-bindings", r.roleBindingsList)
+		v2.GET("/resources/:id/role-bindings/:rb_id", r.roleBindingGet)
```
jnschaeffer, line 90: The resource parameter is redundant here, IMO.

```suggestion
		v2.GET("/role-bindings/:rb_id", r.roleBindingGet)
```

internal/api/rolebindings.go
```diff
@@ -0,0 +1,366 @@
...
+	if err := r.engine.DeleteRoleBinding(ctx, rbRes, resource); err != nil {
+		return r.errorResponse("error updating role-binding", err)
+	}
+
+	resp := deleteRoleBindingResponse{Success: true}
+
+	return c.JSON(http.StatusOK, resp)
+}
+
+func (r *Router) roleBindingGet(c echo.Context) error {
```
jnschaeffer, line 226: As written, this doesn't actually enforce that the role ID and binding ID are related in any way. It would probably make more sense to just use the binding ID and do authZ checks accordingly.

internal/api/rolebindings.go
```diff
@@ -0,0 +1,366 @@
...
+			Subjects: resourceToSubject(rb.Subjects),
+			Role: roleBindingResponseRole{
+				ID:   rb.Role.ID,
+				Name: rb.Role.Name,
+			},
+		},
+	)
+}
+
+func (r *Router) roleBindingUpdate(c echo.Context) error {
```
jnschaeffer, line 286: Same issue as with `roleBindingGet`.

internal/api/router.go
```diff
@@ -86,6 +86,12 @@ func (r *Router) Routes(rg *echo.Group) {
...
+		v2.POST("/resources/:id/role-bindings", r.roleBindingCreate)
+		v2.DELETE("/resources/:id/role-bindings/:rb_id", r.roleBindingsDelete)
+		v2.PATCH("/resources/:id/role-bindings/:rb_id", r.roleBindingUpdate)
```
jnschaeffer, line 93:
```suggestion
		v2.PATCH("/role-bindings/:rb_id", r.roleBindingUpdate)
```

internal/api/types.go
```diff
@@ -80,3 +80,41 @@ type listRolesV2Role struct {
...
+	Name string          `json:"name"`
+}
+
+type roleBindingSubjectCondition struct{}
+
+type roleBindingSubject struct {
+	ID        gidx.PrefixedID              `json:"id" binding:"required"`
+	Type      string                       `json:"type,omitempty"`
+	Condition *roleBindingSubjectCondition `json:"condition,omitempty"`
+}
```
jnschaeffer, line 97: Let's add conditions once we know what they'll look like. I actually suspect we'll want conditions to be on the binding and not the subject, but we should figure that out.

internal/query/rolebindings.go
```diff
@@ -0,0 +1,624 @@
...
+			span.RecordError(err)
+			span.SetStatus(codes.Error, err.Error())
+
+			return types.RoleBinding{}, err
+		}
+
+		rb.Role = types.Role{
+			ID:   roleID,
+			Name: dbRole.Name,
+		}
```
jnschaeffer, line 89: Shouldn't this be outside of the loop?

internal/query/rolebindings.go
```diff
@@ -0,0 +1,624 @@
...
+
+	if _, err := e.client.WriteRelationships(ctx, &pb.WriteRelationshipsRequest{
+		Updates: append(updates, subjUpdates...),
+	}); err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+
+		return types.RoleBinding{}, err
+	}
+
```
jnschaeffer, line 182: Super important: You'll want to update the zedtoken for the resource that has the role binding so the binding takes effect immediately, otherwise there will be a substantial delay. Use `e.upsertZedToken(ctx, resource, zedToken)`.

internal/query/rolebindings.go
```diff
@@ -0,0 +1,624 @@
...
+	// apply changes
+	if _, err := e.client.WriteRelationships(ctx, &pb.WriteRelationshipsRequest{Updates: updates}); err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+
+		return err
+	}
+
```
jnschaeffer, line 240: Same as above re: `upsertZedToken`.

internal/query/rolebindings.go
```diff
@@ -0,0 +1,624 @@
...
+			if errors.Is(err, ErrRoleBindingNotFound) {
+				// print and record a warning message when there's a grant points
+				// to a role-binding that not longer exists.
+				//
+				// this should not happen in normal circumstances, but it's possible
+				// if some role-binding relationships are deleted directly through
+				// spiceDB
+				err := fmt.Errorf("%w: dangling grant relationship: %s", err, rel.String())
+
+				span.RecordError(err)
```
jnschaeffer, line 292: We should record the error regardless of what kind of error it was, right? Also, we need to call `SetStatus` on the span when an error occurs.

internal/query/rolebindings.go
```diff
@@ -0,0 +1,624 @@
...
+	_, err = e.client.WriteRelationships(ctx, &pb.WriteRelationshipsRequest{Updates: updates})
+	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+
+		return types.RoleBinding{}, err
+	}
+
+	rolebinding.Subjects = subjects
+
```
jnschaeffer, line 396: As above, call `upsertZedToken` with the token returned from `WriteRelationships`.

internal/query/rolebindings.go
```diff
@@ -0,0 +1,624 @@
...
+	for _, err := range errs {
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return bindings, nil
+}
+
+func (e *engine) UpdateRoleBinding(ctx context.Context, rb types.Resource, subjects []types.RoleBindingSubject) (types.RoleBinding, error) {
```
jnschaeffer, line 321: Something we need to think about here is locking to ensure we serialize changes to a role binding, or even a resource. As written, we're set up for a number of potential race condition problems for create, update, and delete of bindings.

jnschaeffer (changes requested):
Looks directionally correct so far. A couple of things stick out:

* For mutations, we need to ensure the ZedToken for affected resources gets populated immediately
* We need some kind of locking mechanism similar to `LockRoleForUpdate`. For mutations on bindings, we will want to lock on both the role and the binding so we don't wind up with situations like trying to create a binding on a role that's being deleted

internal/api/rolebindings.go
```diff
@@ -0,0 +1,366 @@
...
+
+	subjectResource, err := r.currentSubject(c)
+	if err != nil {
+		return err
+	}
+
+	// permissions on role binding actions, similar to roles v1, are granted on the resources
+	if err := r.checkActionWithResponse(ctx, subjectResource, string(iapl.RoleBindingActionCreate), resource); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
bailinhe: This worth some discussions, my thought is that we should allow "IAM admins" to assign any permissions to users under the same OU and its children. 

One can argue that "IAM admin" would essentially be a "super user" or owner and I would agree, "IAM admin" is indeed a role that need to be treated very cautiously, but the same goes with "owner" or "super user". And that's where groups come in to provide the flexibility, a group manager will be able to grant certain permissions to users without being an IAM admin.

internal/query/rolebindings.go
bailinhe: I added a [wrapper function `applyUpdates`](https://github.com/infratographer/permissions-api/pull/249/files#diff-8a06560f98df912ac583f2a3b001ba10320e4adaef92cf805eb703c95fa9de4bR1273-R1289) that runs the `WriteRelationships` and also runs `upsertZedToken` on each of the resources

internal/query/rolebindings.go
```diff
@@ -0,0 +1,844 @@
...
+	add, remove := diff(current, incoming)
+
+	// return if there are no changes
+	if (len(add) + len(remove)) == 0 {
+		return rolebinding, nil
+	}
+
+	// 2. create relationship updates
+	updates := make([]*pb.RelationshipUpdate, len(add)+len(remove))
+	i := 0
```
jnschaeffer, line 499: Why not use `make([]*pb.RelationshipUpdate, 0, len(add)+len(remove))` and then call `append` everywhere? You'll still get only one allocation that way since you're using the `make([]T, length, capacity)` form of `make`.

internal/query/roles_v2.go
```diff
@@ -426,7 +430,10 @@ func (e *engine) DeleteRoleV2(ctx context.Context, roleResource types.Resource)
 		errs = append(errs, err)
 	}
 
-	// 2.c TODO remove all role relationships in role bindings associated with this role
+	// 2.c remove all role relationships in role bindings associated with this role
+	if err := e.deleteRoleBindingsForRole(ctx, roleResource); err != nil {
+		errs = append(errs, err)
+	}
```
jnschaeffer, line 436: Why do this? It seems like it would make more sense to disallow deleting a role unless all of its bindings are deleted first.

internal/query/rolebindings.go
```diff
@@ -0,0 +1,844 @@
...
+		return nil
+	case errors.Is(err, ErrActionNotAssigned):
+		return fmt.Errorf("%w: role: %s is not available for resource: %s", ErrRoleNotFound, role.ID, res.ID)
+	default:
+		return err
+	}
+}
+
+// deleteRoleBindingsForRole deletes all role-binding relationships with a given role.
+func (e *engine) deleteRoleBindingsForRole(ctx context.Context, roleResource types.Resource) error {
```
jnschaeffer, line 609: I already noted this, but I'm not sure this is necessary. It looks like it has the potential to take a lot of time, and hard deleting every binding for a role seems like a pretty risky and destructive action.

In general, stopping users from deleting something that might have cascading effects seems more reasonable.

internal/storage/rolebinding.go
```diff
@@ -0,0 +1,297 @@
...
+	tx, err := getContextTx(ctx)
+	if err != nil {
+		return types.RoleBinding{}, err
+	}
+
+	var rb types.RoleBinding
+
+	err = tx.QueryRowContext(ctx, `
+		INSERT INTO rolebindings (id, resource_id, created_by, updated_by, created_at, updated_at)
+		VALUES ($1, $2, $3, $3, now(), now())
```
jnschaeffer, line 136: Nit: Personally I like calling `time.Now()` or something similar once, rather than many "now" invocations.

jnschaeffer (commented):
More thoughts. Looks better, not sure we need all the complexity around batch deleting role bindings when deleting a role (since I'm not sure we want to support that behavior anyway).

internal/query/rolebindings.go
```diff
@@ -0,0 +1,844 @@
...
+	ctx, span := e.tracer.Start(
+		ctx, "engine.GetRoleBinding",
+		trace.WithAttributes(attribute.Stringer("role_binding_id", roleBinding.ID)),
+	)
+	defer span.End()
+
+	rb, err := e.store.GetRoleBindingByID(ctx, roleBinding.ID)
+	if err != nil {
+		if errors.Is(err, storage.ErrRoleBindingNotFound) {
+			err = fmt.Errorf("%w: role-binding: %s", ErrRoleBindingNotFound, err)
```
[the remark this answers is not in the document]
bailinhe: I think we should, it's consistent with `role` and `role_v2`

internal/query/rolebindings.go
```diff
@@ -0,0 +1,844 @@
...
+			"%w: invalid role-binding resource type: %s",
+			ErrInvalidType, e.rbac.RoleBindingResource.Name,
+		)
+
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+		logRollbackErr(e.logger, e.store.RollbackContext(dbCtx))
+
+		return types.RoleBinding{}, err
+	}
```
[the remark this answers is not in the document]
bailinhe: that's correct, part of the schema validation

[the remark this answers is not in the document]
jnschaeffer: Good to know. Maybe we can knock this out and save some lines of code then?

internal/types/types.go
```diff
@@ -100,7 +95,13 @@ type Relationship struct {
...
-	Subjects []RoleBindingSubject
+	ID         gidx.PrefixedID
+	ResourceID gidx.PrefixedID
+	RoleID     gidx.PrefixedID
+	SubjectIDs []gidx.PrefixedID
+
+	CreatedBy gidx.PrefixedID
+	UpdatedBy gidx.PrefixedID
+	CreatedAt time.Time
+	UpdatedAt time.Time
```
jnschaeffer, line 106: Alright, this API feels more right now.

internal/storage/rolebinding.go
```diff
@@ -0,0 +1,247 @@
...
+		if i > 0 {
+			fmtStr = ", $%d"
+		}
+
+		clause += fmt.Sprintf(fmtStr, i+1)
+		args[i] = id.String()
+	}
+
+	return clause, args
+}
```
jnschaeffer, line 247: Do we still need this?

internal/storage/rolebinding.go
```diff
@@ -0,0 +1,247 @@
...
+
+func (e *engine) ListResourceRoleBindings(ctx context.Context, resourceID gidx.PrefixedID) ([]types.RoleBinding, error) {
+	db, err := getContextDBQuery(ctx, e)
+	if err != nil {
+		return nil, err
+	}
+
+	rows, err := db.QueryContext(ctx, `
+		SELECT id, resource_id, created_by, updated_by, created_at, updated_at
+		FROM rolebindings WHERE resource_id = $1
```
jnschaeffer, line 88: Maybe order by `created_at`?

internal/query/roles_v2.go
```diff
@@ -348,6 +352,33 @@ func (e *engine) DeleteRoleV2(ctx context.Context, roleResource types.Resource)
...
+
+	// reject delete if role is in use
+	if len(bindings) > 0 {
+		err := fmt.Errorf("%w: cannot delete role", ErrDeleteRoleInUse)
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+
+		return err
+	}
```
jnschaeffer, line 380: This looks good, but should probably go after we lock the role.

internal/query/relations.go
```diff
@@ -1231,3 +1231,62 @@ func (e *engine) NewResourceFromIDString(id string) (types.Resource, error) {
...
+		if u == nil {
+			continue
+		}
+
+		var op pb.RelationshipUpdate_Operation
+
+		switch u.Operation {
+		case pb.RelationshipUpdate_OPERATION_CREATE:
+			fallthrough
+		case pb.RelationshipUpdate_OPERATION_TOUCH:
```
jnschaeffer, line 1257:
```suggestion
		case pb.RelationshipUpdate_OPERATION_CREATE, pb.RelationshipUpdate_OPERATION_TOUCH:
```

jnschaeffer (commented):
Looks good! A couple of minor comments before I do some manual testing of this.

internal/storage/rolebinding.go
bailinhe: it also used in `rolev2` where we need to grab a bunch of roles, might as well keep it

PR #254, opened 2024-05-10
Do not traverse hidden policy directories

jnschaeffer:
Traversing the policy directory recursively in Kubernetes results in duplicate policies being read, as Kubernetes ConfigMaps store all files as symlinks to a `..data` directory inside the actual volume.

To mitigate this issue, this PR prevents hidden directories below the top level directory from being visited.

### DIFF
--- a/internal/iapl/policy.go
+++ b/internal/iapl/policy.go
@@ -207,7 +207,9 @@ func LoadPolicyDocumentFromFiles(filePaths ...string) (PolicyDocument, error) {
 	return policyDocument, nil
 }
 
-// LoadPolicyDocumentFromDirectory reads the provided directory path, reads all files in the directory, merges them, and returns a new merged PolicyDocument.
+// LoadPolicyDocumentFromDirectory reads the provided directory path, reads all files in the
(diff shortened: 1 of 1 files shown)


internal/iapl/policy.go
```diff
@@ -208,6 +208,7 @@ func LoadPolicyDocumentFromFiles(filePaths ...string) (PolicyDocument, error) {
 }
 
 // LoadPolicyDocumentFromDirectory reads the provided directory path, reads all files in the directory, merges them, and returns a new merged PolicyDocument.
+// This does not recursively traverse directories.
```
mikemrm, line 211: Recursive directories is something we want to keep to allow for better organization. It might be better to limit the directories traversed by filtering out directories prefixed with `.`.

mikemrm (changes requested):

internal/iapl/policy.go
jnschaeffer: Sounds good. Let me update this PR accordingly.

mikemrm (approved):

PR #256, opened 2024-05-15
Use SpiceDB directly instead of NATS KV for ZedToken updates

jnschaeffer:
NATS introduces an additional dependency for getting ZedToken updates that we don't really need if we can just use the SpiceDB Watch API. This PR updates the permissions-api server to use SpiceDB Watch instead for ZedToken updates.


internal/query/zedtokens.go
```diff
@@ -19,48 +15,49 @@ const (
...
+			relationship.Subject.Object.ObjectId,
+			zedToken,
+		)
 	}
+}
 
-	ttl := status.TTL()
+// initZedTokenCache creates a new LRU cache that watches SpiceDB for ZedToken updates.
+func (e *engine) initZedTokenCache(ctx context.Context) error {
+	ttl := time.Minute
```
jnschaeffer, line 38: Note to self: we should make this configurable based on the revision quantization interval.

internal/query/zedtokens.go
```diff
@@ -19,48 +15,49 @@ const (
...
-			if entry == nil {
-				continue
-			}
-
-			key := entry.Key()
-			value := string(entry.Value())
+		for {
+			resp, err := watchClient.Recv()
+			if err != nil {
+				e.logger.Errorf("error receiving updates", "error", err)
```
mikemrm, line 53: should we exit this goroutine if the error is a context cancelled?

mikemrm (approved):
looks good

bailinhe (commented):
looks like one of the tests need to be updated

https://github.com/infratographer/permissions-api/actions/runs/9097856914/job/25006681107?pr=256#step:3:2763

jnschaeffer:
Closing in favor of #257.

PR #257, opened 2024-05-15
Add design proposal and implementation for ZedTokens table

jnschaeffer:
While we have some guarantees that we try to make around consistency in permissions-api, particularly around ZedTokens, these were not explicitly spelled out anywhere. This PR adds a design proposal for using CRDB itself to store ZedTokens for resources, rather than relying on streaming from SpiceDB or lookups in NATS KV. It also adds the implementation, since the work involved is pretty minimal and demonstrating that this approach is workable is good, IMO.

The motivations for the CRDB-based approach are that it grants us the consistency guarantees spelled out in the proposal while also allowing for more flexibility than NATS provides around the location of a given leader for coordinating reads. Using CRDB also removes a dependency from the permissions-api critical path.


mikemrm (dismissed):
looks good!

docs/20240515-zedtokens.md
```diff
@@ -0,0 +1,46 @@
...
+* If a ZedToken exists for that resource, use `at_least_as_fresh` with the given ZedToken
+* If no ZedToken exists, use `minimize_latency`
+
+The advantage of this approach is that it keeps management of consistency within permissions-api, meaning future changes do not (necessarily) require updates to other services, nor does it change current API semantics. Additionally, it does not introduce new dependencies to permissions-api, instead leveraging CRDB's availability and fault tolerance guarantees.
+
+## Constraints and limitations
+
+As designed, this only provides immediately visible updates for a given resource; other indirectly affected resources are not updated (but could be in the future). Additionally, given CRDB leaseholder behavior, multi-region deployments may see longer durations when performing permissions checks if the ZedToken table is not a global table.
+
+## Alternatives considered
```
senior, line 36: I'm still trying to understand this full stack, so excuse my ignorance. It seems like another alternative is no ZedTokens at all. Reading [this post](https://authzed.com/blog/prevent-newenemy-cockroachdb) seems to indicate that this won't happen in CRDB. Also it seems even less likely because of our [prefixing](https://github.com/infratographer/permissions-api/blob/main/docs/iapl.md#mapping-policies-to-spicedb).

Am I missing something in my reading of that post and our usage of SpiceDB?

jnschaeffer: So that blog post is primarily about controlling write ordering in CRDB, not ACL updates. There are a few strategies we can adopt if the current default of using a single key for ordering transactions becomes a problem, but we haven't run into scaling issues there.

Where things get more interesting is around ZedTokens. The use of ZedTokens isn't really specific to CRDB, but rather inherent to how Zanzibar and SpiceDB operate. SpiceDB revisions (known as snapshots in the Zanzibar paper) are generally some number of seconds long and reads happen within a given revision. All datastores, including CRDB, use revisions for reading data.

ZedTokens allow for the client to request data newer than whatever a given SpiceDB replica might have available in its cache or whatever the current revision it's using is. In a CRDB deployment environment that uses follower reads, this would mean requesting data that might be newer than what the nearest replica has, thus incurring a one-time round trip to the leaseholder coordinating work. Following this, the data are cached in memory in SpiceDB for the remainder of the revision duration. Other storage engines don't have the same horizontal/multi-region scaling capability that CRDB offers.

Not using ZedTokens would require either waiting for a new revision to elapse before ACL updates are visible or leveraging [full consistency][consistency] in permissions checks, which means we get none of the caching benefits SpiceDB provides.

[consistency]: https://authzed.com/docs/spicedb/concepts/consistency#levels

mikemrm (approved):

PR #260, opened 2024-06-05
Add idprefix validation

bailinhe:
Add `idprefix` validation on IAPL validate

### DIFF
--- a/internal/iapl/policy.go
+++ b/internal/iapl/policy.go
@@ -11,6 +11,7 @@ import (
 
 	"go.infratographer.com/permissions-api/internal/types"
 
+	"go.infratographer.com/x/gidx"
 	"gopkg.in/yaml.v3"
 )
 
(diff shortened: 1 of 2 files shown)


internal/iapl/policy.go
```diff
@@ -289,10 +289,15 @@ func (v *policy) validateResourceTypes() error {
...
 			for _, tt := range rel.TargetTypes {
-				if _, ok := v.rt[tt.Name]; !ok {
+				rr, ok := v.rt[tt.Name]
+				if !ok {
 					return fmt.Errorf("%s: relationships: %s: %w", resourceType.Name, tt.Name, ErrorUnknownType)
 				}
 
+				if _, err := gidx.NewID(rr.IDPrefix); err != nil {
+					return fmt.Errorf("%s: relationships: %s: %w", resourceType.Name, tt.Name, err)
+				}
```
mikemrm, line 299: This seems to be testing the prefix on every related resource type instead of just the individual resource type. If a resource was created with no other resources relating to it, it's idprefix would not be checked.

I believe using `resourceType` from line 289 just below it might be the proper place to do that validation.

mikemrm (changes requested):

mikemrm (approved):

PR #290, opened 2024-09-26
Don't defaulty inherit all actions for role binding v2

mikemrm:
By default when specifying `InheritPermissionsFrom` on resource type's RoleBindingV2 all actions were being inherited rather than only inheriting RBAC v2 actions.

This change reduces the default inheritance to only the RBAC v2 actions but setting `InheritAllActions` to `true` will automatically include inheritance for all configured actions on that resource type.


bailinhe (approved):
image

PR #291, opened 2024-09-27
role-binding: remove at least one subject requirement

bailinhe:
Remove at least one subject as hard requirement for creating new role-bindings.


mikemrm (commented):
It doesn't look like we have tests for updating a role binding to have no subjects. Can we add those tests and to make sure that works as well?

mikemrm (approved):

PR #292, opened 2024-10-02
add manager field to simplify multi service management

mikemrm:
This adds a manager field to roles and role bindings allowing services to mark the manager of a specific role / role binding.

This allows the managing service to easily filter out unmanaged roles / role bindings which it does not own which allows for cleaner logic for reconciling roles/role bindings.

Both roles and role binding create requests can provide an optional `manager` field and role / role binding list endpoints can now be provided an optional `manager` query value.


bailinhe (approved):
🚀

PR #294, opened 2024-10-23
Add Bulk Permission Checks Endpoint

bailinhe:
`POST /v1/allow/bulk`

This endpoint takes a list of `<resource, action>` as input, and returns the permission checks outcome of each input

### DIFF
--- a/internal/api/permissions.go
+++ b/internal/api/permissions.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/labstack/echo/v4"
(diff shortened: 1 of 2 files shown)


internal/api/permissions.go
```diff
@@ -129,6 +130,17 @@ type checkResult struct {
 	Error   error
 }
 
+type bulkCheckActionsRequest []checkAction
+
+type checkActionResponse struct {
+	ResourceID    string `json:"resource_id"`
+	Action        string `json:"action"`
+	HasPermission bool   `json:"has_permission"`
```
jacobsee, line 138: I think I have a _minor_ preference for a boolean result to be named something like `Allowed` given that we have some consumer-facing stuff that refers to access check results as "allowed" or "denied" elsewhere (like https://github.com/metal-toolbox/iam-runtime/blob/main/proto/authorization/authorization.proto#L40-L44) but otherwise 👍

jacobsee (approved):

internal/api/permissions.go
mikemrm: I agree, we should keep similar terminology.

mikemrm (changes requested):

mikemrm (approved):

PR #306, opened 2024-12-17
Create driver and migrations for spanner (psql)

bailinhe:
- [x] create seperate migration directories for spanner SQL files
- [x] create `pgx` driver to use spanner-pg-adapter
- [x] create additional config directory to configure `pgx` driver
- [x] create docker compose file for building dev containers with
  spanener and spanner-pg-adapter

Signed-off-by: Bailin He <bahe@equinix.com>


mikemrm:
Also missing are helm chart updates for these new configuration options.

chart/permissions-api/templates/deployment-worker.yaml
```diff
@@ -55,13 +55,24 @@ spec:
           env:
             - name: PERMISSIONSAPI_SERVER_LISTEN
               value: ":{{ include "permapi.listenPort" . }}"
+          {{- if eq .Values.config.db.driver "postgres"}}
```
mikemrm, line 58: Do we need the conditional? Why not just rely on the `with`, similar to how crdb was doing before?

`Values.config.db.driver` should be `Values.config.db.engine`

mikemrm (changes requested):

mikemrm (dismissed):

mikemrm (approved):

